PR #186, opened 2014-04-15
Allow #merge! to accept Jbuilder objects

johnkchow:
The documentation didn't have any example on how to use jbuilders within an `#array!` block call, so after digging around the source code I realize that there's a `#merge!` method that I can use. This change makes the code cleaner when trying to use Jbuilder objects within the array! block.

Original code would look like:

``` ruby
json.array! @people do |person|
  json.merge! person.to_builder.attributes!
end
```

New code would look like:

``` ruby
json.array! @people do |person|
  json.merge! person.to_builder
end
```

I was tempted to not even include the `json.merge!` and have the block be smart about what object is returned, but I figure this is consistent with the rest of the API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,17 @@ end
 # => [ { "name": "David", "age": 32 }, { "name": "Jamie", "age": 31 } ]
 ```
 
+You can merge Jbuilder objects within the array block
+
+``` ruby
+# @people = People.all
+json.array! @people do |person|
+  json.merge! people.to_builder
+end
+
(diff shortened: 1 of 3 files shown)


krevels:
@rwz, I know this issue is closed for some time now, but I'm unclear as to why it was suggested that the model implement `as_json` if it's not what you recommend.  Doesn't this PR present a better alternative to doing just that? 

The advantage I see here to the approach @johnkchow uses is that it allows for the composition of jbuilder objects within the view template in the same way that you would compose them within object's `to_builder` method, rather than having to convert them to a Hash first.

Also, it would be nice to just have `merge!` documented somewhere in the README

PR #251, opened 2015-02-23
partial! + :locals

amatsuda:
I was surprised to see that `partial!` doesn't handle `:locals` option in the same way as ActionView.

There was a comment in code that implies `partial! partial: 'name', locals: { foo: 'bar' }` would work, but in fact that code does not work, and does not even tested.

So here's an implementation with tests for:
`json.partial! 'partial_name', locals: {foo: 'bar}`
and
`json.partial! 'partial_name', foo: 'bar`

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,17 @@ json.partial! partial: 'posts/post', collection: @posts, as: :post
 json.comments @post.comments, partial: 'comment/comment', as: :comment
 ```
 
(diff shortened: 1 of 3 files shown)


rwz:
Sweet. Thanks for PR. Will release a new version shortly.

amatsuda:
:cool:

rwz:
Released 2.2.9

amatsuda:
Thanks! :)

PR #256, opened 2015-03-05
Add #collect!

cheald:
This adds `#collect!(key)`, which is useful when you would want to build an array without knowing the inputs for it ahead of time. I'm using this for building Elasticsearch queries, like:

    Jbuilder.encode do |j|
      j.query { j.match_all }
      j.collect!(:filter) do
        j.range(a: {to: 3})  if condition?("a")
        j.range(b: {to: 5})  if condition?("b")
        j.range(c: {to: 10}) if condition?("c")
      end
    end

Rather than the less pleasant:

    Jbuilder.encode do |j|
      j.query { j.match_all }
      j.filter do
        j.child! { j.range(a: {to: 3})  } if condition?("a")
        j.child! { j.range(b: {to: 5})  } if condition?("b")
        j.child! { j.range(c: {to: 10}) } if condition?("c")
      end
    end

When collecting, each child clause is collected as its own sub-document into the parent array.

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -25,38 +25,42 @@ class Jbuilder
   BLANK = Blank.new
 
   def set!(key, value = BLANK, *args, &block)
-    result = if block
-      if !_blank?(value)
-        # json.comments @post.comments { |comment| ... }
-        # { "comments": [ { ... }, { ... } ] }
-        _scope{ array! value, &block }
-      else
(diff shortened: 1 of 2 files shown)


rwz:
This feature changes the behavior and makes it really indirect and confusing. I don't think it'll be a good addition to existing API and will be useful for other users.

In this particular case I'd extract filter into a helper method that'll just return a pre-formatted array.

PR #257, opened 2015-03-06
Add Blank#empty? to tell if attributes are empty or not

a2ikm:
Before:

```ruby
json.foo do
  json.attributes!.empty? #=> false
end
````

After:

```ruby
json.foo do
  json.attributes!.empty? #=> true
end
````

In my case, I want to ensure creating an array like:

```ruby
json.foo do
  if some_condition?
    json.child! do
      # Build a hash
    end
  end

  # and other conditional child! calling....

  # Ensure creating array if no child! called above.
  if json.attributes!.empty?
    json.array! []
  end
end
```

This worked with v2.1.2, but not with v2.2.9 because  `json.attributes!` returns `Jbuilder::Blank` object if attributes are empty.

### DIFF
--- a/lib/jbuilder/blank.rb
+++ b/lib/jbuilder/blank.rb
@@ -3,5 +3,9 @@ class Jbuilder
     def ==(other)
       super || Blank === other
     end
+
+    def empty?
+      true
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


rwz:
`json.attributes!.empty?` feels a little janky, but I agree, Blank object should totally respond to `empty?`, `present?` and `blank?` properly!

a2ikm:
Thank you!

PR #258, opened 2015-03-07
Why don't we generate jbuilder templates for Rails 5?

amatsuda:
The patch has no test, but I confirmed that it works on my local rails/master app.

### DIFF
--- a/lib/jbuilder/railtie.rb
+++ b/lib/jbuilder/railtie.rb
@@ -10,7 +10,7 @@ class Jbuilder
       end
     end
 
-    if Rails::VERSION::MAJOR == 4
+    if Rails::VERSION::MAJOR >= 4
       generators do |app|
         Rails::Generators.configure! app.config.generators
         Rails::Generators.hidden_namespaces.uniq!


rwz:
Well, Jbuilder also depends on activesupport "< 5". I'm not sure I'm comfortable announcing that Rails 5 is supported just yet, and that PR kind of does that. Could we at wait until they at least release a beta?

rwz:
I'll leave this PR open and merge it to give credit once Rails 5 beta is available.

amatsuda:
@rwz Thank you for the consideration, but `activesupport "< 5"` would not really be a problem in this case.
Because rails master is already versioned as `5.0.0.alpha` 
https://github.com/rails/rails/blob/ed9589ee62c3e19971d620aeab9c7e811cf154c4/RAILS_VERSION 
which will actually match both `<5` and `>= 4` thanks to RubyGems' cleverness.

```ruby
Gem::Version.new('5.0.0.alpha') < Gem::Version.new('5')
#=> true
```

rwz:
Alright, that makes sense. :)

rwz:
Released 2.2.11

amatsuda:
@rwz Thank you! :star2:

PR #260, opened 2015-04-03
remove splats

ASnow:
Removed +arguments+ splat operator in private method +#_extract_hash_values+
Removed +arguments+ splat operator in private method +#_extract_method_values+

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -207,9 +207,9 @@ class Jbuilder
   #   json.(@person, :name, :age)
   def extract!(object, *attributes)
     if ::Hash === object
-      _extract_hash_values(object, *attributes)
+      _extract_hash_values(object, attributes)
     else
-      _extract_method_values(object, *attributes)
+      _extract_method_values(object, attributes)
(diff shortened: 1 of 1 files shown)


rwz:
What's the purpose of this change?

ASnow:
Performance optimization
```
require 'benchmark'

class MethodWithSplat
  def method1 a1, *args
    method2 a1, *args
  end
  private
  def method2 a1, *args
  end
end
class MethodWithoutSplat
  def method1 a1, *args
    method2 a1, args
  end
  private
  def method2 a1, args
  end
end


a = MethodWithSplat.new
b = MethodWithoutSplat.new

called = 10_000_000
Benchmark.benchmark do |x|
  x.report("with splat:")   { called.times{ a.method1 nil, nil, nil } }
  x.report("without splat:")   { called.times{ b.method1 nil, nil, nil } }
end

#with splat:  6.030000   0.000000   6.030000 (  4.133987)
#without splat:  0.000000   0.000000   0.000000 (  2.446851)
```

PR #273, opened 2015-06-08
Fix scaffold controller generator rails api

jmbejar:
It generates an API Controller when the host rails app is generated with the --api option.

This PR does not make sense until github.com/rails/rails/pull/19832 is merged.

As a bonus, I added a new configuration in Appraisal to test against rails master.

/cc @spastorino


rwz:
LGTM

test/scaffold_api_controller_generator_test.rb
```diff
@@ -0,0 +1,46 @@
+require 'test_helper'
+require 'rails/generators/test_case'
+require 'generators/rails/scaffold_controller_generator'
+
+if Rails::VERSION::MAJOR >= 5
```
rwz: Could this be just `> 4` instead?

jmbejar: :+1: Done!

Issue #270, opened 2015-05-26
array index


khoan:
Is there no way to get the index of the array item? 

`json.array_with_index! @people do |person,index|` would be nice.

tchret:
+1

paritosh90:
+1

wrmk:
> Just use Ruby:
> 
> json.array! @people.each_with_index.to_a do |(person, index)|
>   json.index index
>   json.name person.name
> end


I suppose that it makes twice iteration on `@people`

PR #279, opened 2015-06-28
Add #array_with_index! method

kamilbielawski:
Requested in #270


yamadayuki:
:+1:

rwz:
I recommend using pure Ruby for that instead of adding more functionality to Jbuilder.

```ruby
json.array! @people.each_with_index.to_a do |(person, index)|
  json.index index
  json.name person.name
end
```

Issue #277, opened 2015-06-22
jbuilder local variables do not work with jruby since 2.3.0


badazz91:
Hi guys,

when I pass a local variable to a partial inside of a json.array block, it is not accessible in the partial.
The strange thing is, that this only occurs when using jruby and jbuilder 2.3.0. If I switch back to 2.2.16 it also works with jruby.
With mri-ruby it also works with 2.3.0.

rwz:
Can you provide a code example?

sytzeloor:
I encountered the same issue. Here is a code sample:

```
json.ranking(
  @course.students,
  partial: 'api/v1/students/rank',
  as: :student,
  student_id: @student.id
)
```

The variable `student_id` is not available in the `rank` partial.

rwz:
Well, it's not supposed to work this way. Try this:

```ruby
json.ranking @course.students,
  partial: "api/v1/students/rank",
  as: :student,
  locals: { student_id: @student.id }
```

fpauser:
@sytzeloor Why ruby-version are you using?

PR #290, opened 2015-08-24
fixed documentation due to issue #277

badazz91:
Like discussed in #277 locals have to be explicitly passed to a partial.


rwz:
Well, this is incorrect. This syntax still works fine.

PR #292, opened 2015-09-01
Add CONTRIBUTING.md

maclover7:
cc @rafaelfranca

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,5 @@
 # Jbuilder
 
-[![Build Status](https://api.travis-ci.org/rails/jbuilder.svg)][travis]
-[![Gem Version](http://img.shields.io/gem/v/jbuilder.svg)][gem]
-[![Code Climate](http://img.shields.io/codeclimate/github/rails/jbuilder.svg)][codeclimate]
-[![Dependencies Status](http://img.shields.io/gemnasium/rails/jbuilder.svg)][gemnasium]
-
-[travis]: https://travis-ci.org/rails/jbuilder
-[gem]: https://rubygems.org/gems/jbuilder
-[codeclimate]: https://codeclimate.com/github/rails/jbuilder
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,15 +1,5 @@
...
-[![Build Status](https://api.travis-ci.org/rails/jbuilder.svg)][travis]
-[![Gem Version](http://img.shields.io/gem/v/jbuilder.svg)][gem]
-[![Code Climate](http://img.shields.io/codeclimate/github/rails/jbuilder.svg)][codeclimate]
-[![Dependencies Status](http://img.shields.io/gemnasium/rails/jbuilder.svg)][gemnasium]
-
-[travis]: https://travis-ci.org/rails/jbuilder
-[gem]: https://rubygems.org/gems/jbuilder
-[codeclimate]: https://codeclimate.com/github/rails/jbuilder
-[gemnasium]: https://gemnasium.com/rails/jbuilder
-
```
rwz: Not sure I'm fully onboard with removing badges from README. What's the rationale behind this?

rafaelfranca: Mainly because they are not useful for the final users.

* Travis badge don't add anything since we expect the build to be always green.
* Rubygems badge only give the last release, they can get the same information form GItHub release page.
* Codeclimate badge I don't even think codeclimate is useful, but even if it is gem users don't need to care about it
* Gemnasium it has useful informations only for contributors.

So, IMHO all badges are not useful for the users of the library but they may be useful for contributors so I think it belongs to the CONTRIBUTING.md file.

rwz: Well, ¯\\\_(ツ)_/¯

Alright then

rafaelfranca: Also https://github.com/lostisland/faraday/commit/ac1b537f49bf110e87fe50679d80764367fb2d13 contains good reasons.

PR #295, opened 2015-09-28
Add #extract_without_nil!

nambrot:
At @WellframeInc, we have a very outspoken member that advocates for not passing a key where the value is `null`. We couldn't actually find a consensus online as to what is the right convention. This PR adds the ability to specify `extract!` but to omit keys with `extract_without_nil!`.

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -220,6 +220,14 @@ class Jbuilder
     end
   end
 
+  def extract_without_nil!(object, *attributes)
+    if ::Hash === object
+      _extract_hash_values(object, attributes, false)
+    else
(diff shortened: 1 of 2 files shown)


rwz:
Isn't `ignore_nil` options is exactly what you're looking for in this case? 

https://github.com/rails/jbuilder/blob/v2.3.1/lib/jbuilder.rb#L16

Not sure why it's undocumented though :(

nambrot:
Nvm then!

rwz:
Need to probably invest into some documentation around it.

nambrot:
https://github.com/rails/jbuilder/pull/296

PR #296, opened 2015-09-28
Add `ignore_nil! to documentation`

nambrot:
@rwz

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,15 @@ json.author do
 end
 ```
 
+To prevent Jbuilder from including null values in the output, you can use the `ignore_nil!` method:
+
+```ruby
+json.ignore_nil!
+json.foo nil
+json.bar "bar"
+# => { "bar": "bar" }
+```
+
 Fragment caching is supported, it uses `Rails.cache` and works like caching in
 HTML templates:
 


nambrot:
Any idea why this would fail?

README.md
```diff
@@ -196,6 +196,14 @@ json.author do
 end
 ```
 
+Some languages/frameworks don't handle `null` well, in which case you can use `ignore_nil!`
```
jeremy: Specific examples? This raises more questions than it answers.

nambrot: Is this better? https://github.com/rails/jbuilder/commit/1268ef4e0763ea72a0ffb1d57a1161eecd2ef20b

rwz:
@nambrot what if we don't mention other frameworks or languages at all? That seems secondary to the subject of the documentation. Something like this might work:

> To prevent Jbuilder from including null values in the output, you could use `ignore_nil!` method:

```ruby
json.ignore_nil!
json.foo nil
json.bar "bar"
# => { "bar": "bar" }
```

nambrot:
Updated?

rwz:
LGTM. Could you rebase and squash the commits into one?

nambrot:
Did it

rwz:
Thanks. :+1:

Issue #289, opened 2015-08-22
Proposal for caching improvements


vincentwoo:
Regarding the issues brought up in #259...

### Motivation

We'd like to make caching in JBuilder useful. Currently, caching does not offer significant performance benefits for pretty much any app. This is largely due to the serialization and over-the-wire costs JBuilder pays when storing and fetching cached content, because the current serialization scheme doesnt allow for raw partial content. Below is a proposal for enabling much faster caching while still retaining MultiJson as the serialization backend:

### Proposal

1. When `json.cache! cache_key, &block` is encountered in a template, Jbuilder emits a placeholder UUID key/value pair into the document (like `"jbuilder-XXXX": null`) and moves on. It stores the (template digested) cache keys in a list of keys to be looked up, along with their corresponding UUID.

2. After the template is rendered, the half-baked string is held in memory while JBuilder sends a `fetch_multi` for each of the keys in the set.

3. For each key that is found in the cache, the corresponding UUID is blindly replaced with the value taken from the cache. In this case, a simple gsub for `"jbuilder-XXXX": null`, replacing it with whatever the cache has, like: `"color": "red", "count": 2`

4. For each key that is NOT found, execute the block with a fresh `json` object. Once completed, serialize that `json` object to a string, and strip off the leading and trailing braces. Store that resulting string under the cache key and perform the substitution operation described in (3). The brace stripping must happen because the `json.cache!` call does not by default happen under a named key, and can be embedded in the a flow of a template, e.g.

  ```ruby
  json.key1 'value1'
  json.cache! do
    json.key2 'value2'
  end
  ```

  produces `{ "key1": "value1", "key2": "value2"}`  

  Rendering `json.key2 'value2'` in isolation produces `{"key2": "value2"}`, which we then strip of braces, which renders it fit to be inserted into the flow of other JSON objects.

### Cons

* This is a giant fucking hack.
* Currently, the following case works fine in that only one entry is emitted:

  ```ruby
  json.key 'initial value'
  json.cache! do
    json.key 'new value'
  end
  ``` 

  But under the described proposal, JBuilder could double-emit entries for `key`. By and large this will probably not matter - there's no real reason to do this, and most JSON parsers assume the latest definition is takes precedence anyhow.

* Stuff starts to get tricky with nested partials/caches. It should all work fine in theory, there's just going to be a bookkeeping overhead.

dhh:
[...]

PR #298, opened 2015-10-06
Caching improvements

dirk:
Implementation of the serialization-deserialization-overhead-eliminating strategy outlined in #289. Please see that issue for an outline and general details of the strategy.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake"
+gem "rails"
 gem "mocha", require: false
 gem "appraisal"
 gem "pry"
--- a/Rakefile
+++ b/Rakefile
@@ -2,22 +2,22 @@ require "bundler/setup"
 require "bundler/gem_tasks"
 require "rake/testtask"
 
+Rake::TestTask.new do |test|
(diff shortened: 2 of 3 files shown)


dirk:
Please see [this comment](https://github.com/rails/jbuilder/issues/289#issuecomment-146000448) on the proposal pull request for performance results.

Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gemspec
 
 gem "rake"
+gem "rails"
```
peteygao: Is this an oversight? We shouldn't need to bring in the entire Rails gem into jbuilder, right?

rromanchuk:
@dirk would love to try this out, do you have a fork you are maintaining for your own projects?

rromanchuk:
have you seen before? 

`ActionView::Template::Error (can't add a new key into hash during iteration):`  occurs when there is another nested `cache!`

mactkg:
@rromanchuk I think the error is raised when a template call same template itself.

```rb
# model/_template.jbuilder
if model[:child].present?
  json.child model[:child], partial: 'model/template', as: :model
else
  json.child []
end
```

PR #301, opened 2015-10-22
merge! working with nested jbuilder object

bzanchet:
...just like set! and the call syntax.

Hey there!

Here's the need: I like having "presenter" objects when implementing an API in order to separate interface and implementation - they all implement `to_builder`. I have an endpoint that renders a list of objects. Since JSON doesn't allow for top-level arrays, I end up writing template code like:

```
json.objects @objects do |object|
  json.merge! object.to_builder.attributes!
end
```

because `merge!` doesn't take a `Jbuilder` object.

This patch makes it possible to rewrite the previous code as:

```
json.objects @objects do |object|
  json.merge! object.to_builder
end
```

Not a big change, I think..

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -268,7 +268,9 @@ class Jbuilder
   end
 
   def _merge_values(current_value, updates)
-    if _blank?(updates)
+    if ::Jbuilder === updates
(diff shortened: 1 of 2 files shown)


bzanchet:
Looks like `array!` now works too.. :)

Issue #307, opened 2015-12-21
Compatibility with Rails 5.0


mvastola:
With the release of Rails 5.0.0.beta1 just days ago, many users (myself included) are seeking to upgrade their current Rails 4.2 apps to ensure forward-compatibility with this new version of Rails and to make use of its new features and capabilities.

Currently, the `jbuilder.gemspec` includes the line:
```
s.add_dependency 'activesupport', '>= 3.0.0', '< 5'
```
which prevents usage of this gem (or even experimentation with it) in this new version of Rails.

I would like to request that jbuilder be made compatible with Rails 5.0.

rwz:
Mind sending a PR?

mvastola:
I'm more than happy to change the gemspec and run the test suite and see if everything passes. 
Unfortunately, I'm not very familiar with the internals of this gem (nor Rails 5), so I might not be the best person to troubleshoot any compatibility issues with the new Rails.

Stand by for updates.

matthewd:
FYI, I don't recall any deprecation-less changes in AS, so it's probably fine.

mvastola:
So clearly I didn't get to this last night. I'll try again for tonight I'm UTC-5, but people should feel free to beat me to it if they so desire!
On Dec 21, 2015 12:34 AM, Matthew Draper <notifications@github.com> wrote:FYI, I don't recall any deprecation-less changes in AS, so it's probably fine.

mvastola:
Update: So I'm afraid I'm running into difficulties getting the appraisals gem to work correctly. It might be best if someone with a deeper knowledge of jbuilder gives this a go. (I had assumed I could just change the dependency, run `bundle update` and `bundle exec rake test`, but that seems to not be the case -- or at least when I attempt this I get no output.)

NB: I did manage to -- after toying with removing appraisal entirely and then editing Gemfile (on my local copy) to include Rails 5.0 / Arel 7.0 / Rack 2.0 (which no longer need to fetch from github) -- run the included tests and see them all pass.. So all signs point to jbuilder not needing any actual changes. I just don't feel comfortable -- from my present understanding of appraisal -- generating a PR that's inclusive of all the changes needed.

mvastola:
FYI, I figured this out, so PR #309 has been made.

mvastola:
PR #309 has been merged. Closing.

PR #309, opened 2015-12-27
Enabled interoperability with Rails 5.0.x...

mvastola:
Created relevant Appraisal and configured Travis CI testing for Rails 5.0

All CI tests passing with all Rails versions.

Resolves issue #307.


.travis.yml
```diff
@@ -7,7 +7,7 @@ rvm:
   - 1.9
   - 2.0
   - 2.1
-  - 2.2
+  - 2.2.2
```
rwz: why change this?

mvastola: Rails 5.0.x has a minimum ruby version requirement of 2.2.2. Specifying just `2.2` to Travis resolves as version 2.2.0 and causes bundler to fail.

rwz: That's weird, it should resolve to the latest 2.2.x according to their specs.

mvastola: It's an RVM thing, I believe.. see [here](https://github.com/rvm/rvm/blob/stable/config/known#L10). Specifying just `2.2` resolves to `2.2.1` (I was apparently mistaken above when I said `2.2.0`) in the latest stable release of RVM. (It seems the revision number will jump on the next RVM release as this is different in RVM's `master` branch.)

In any case, I assure you it's necessary. I only changed it after it didn't work.

mvastola: Also see https://travis-ci.org/rails/jbuilder/jobs/83196945#L90, the latest CI test of `jbuilder` `master` branch where the RVM ruby version was specified as `2.2`.

rwz:
Thanks for the PR. LGTM.

I've [tweeted at TravisCI](https://twitter.com/rwz/status/681017712307380224), let's see if we could make them fix the ruby 2.2 problem before merging this.

mvastola:
Ok. 

FYI though, looking at https://travis-ci.org/rails/jbuilder/jobs/83196945#L86, it would seem that the issue isn't really with Travis since Travis is passing the version info correctly to RVM. And then RVM is simply interpreting the revision number as documented. (If you were to run `rvm list known` the then-current release of RVM would have had the entry `[ruby-]2.2[.0]`.)  

If you want to work around this, you can add this to `.travis.yml` (since, as I pointed out, its `master` branch, has `2.2` resolving to `2.2.4`), but note that you will then be relying on a dev version of RVM.
```
before_script:
  - rvm get master
```

The alternative is to just wait for the next version of RVM to come out and then you should be able to switch `2.2.2` back to `2.2` in `.travis.yml`.

I'd be surprised if Travis tweets you back anything to the contrary, but can't hurt to see what they have to say, I suppose..

Do you happen to recall where in the Travis docs you saw the behavior specified that the latest version should be pulled?

rwz:
> Do you happen to recall where in the Travis docs you saw the behavior specified that the latest revision of ruby matching the major/minor version specified would be used?

Nope.

I guess you're right, my latest stable RVM points 2.2 to 2.2.1, which is better than 2.2.0 but still not good enough.

Sigh, let me try the `rvm get master` hack.

rwz:
Alright, the `get master` hack didn't work. I've specified full versions for 2.2 and 2.3 and it [seems to be working fine](https://travis-ci.org/rails/jbuilder/builds/98948068). Could you please rebase your PR against master so I could merge it?

mvastola:
Hmm. Yea. It looks like `before_script` doesn't run until way after the ruby is installed.
Rebasing now.

rwz:
Awesome, thanks

mvastola:
Thank you as well!

rwz:
Released 2.4.0

PR #318, opened 2016-02-08
Fix generation of broken controller on Rails 5.0 when ran w/o fields

diego-silva:
This PR fixes the issue reported on rails/rails#23314 that was already fixed on Rails default controller scaffold with PR rails/rails#22574

As stated on the Rails issue and PR, without this fix a newly generated controller will raise an `ArgumentError` on Rails 5.0:
```ruby
ArgumentError: When assigning attributes, you must pass a hash as an argument.
    app/controllers/carts_controller.rb:44:in `block in update'
    app/controllers/carts_controller.rb:43:in `update'
    test/controllers/carts_controller_test.rb:37:in `block in <class:CartsControllerTest>'
```

I believe that since it was changed on the Rails repo we should try to keep the same behaviour on JBuilder, as JBuilder is included by default with Rails and thus always overwrites the controller scaffold for new apps.


rwz:
LGTM. Thanks.

PR #327, opened 2016-03-13
Fix ActionView::Rendering include order with jbuilder

meinac:
Follow up to rails/rails#24178

### DIFF
--- a/lib/jbuilder/railtie.rb
+++ b/lib/jbuilder/railtie.rb
@@ -11,7 +11,6 @@ class Jbuilder
 
       if app.config.respond_to?(:api_only) && app.config.api_only
         ActiveSupport.on_load :action_controller do
-          include ActionView::Rendering
           include ActionController::Helpers
           include ActionController::ImplicitRender
         end


jeremy:
Will this break jbuilder on Rails <5?

maclover7:
@jeremy --> based on [the Travis run](https://travis-ci.org/rails/jbuilder/builds/115649530), it looks like the change passes for all Rails versions tested (>= 3.0)

jeremy:
It's troubling that removing this module doesn't make tests fail, though!

Why was it ever included?

What changes when it's omitted?

What tests should demonstrate that?

maclover7:
Looks like this was added when rails-api was merged into Rails proper -- https://github.com/rails/jbuilder/commit/29c0014a9c954c990075d42c45c66075260e924b -- https://github.com/rails/rails/pull/19832#issuecomment-102153928.

morgoth:
I guess this can be closed after https://github.com/rails/jbuilder/commit/64c510ec69d9e63b73ffd5942e802d21a7d14701

jmbejar:
We have now a fix for the ActionView::Rendering include order issue, although some regression testing is still pending. Please see https://github.com/rails/jbuilder/compare/fcc649046d8b01f3f05cdde7c665ce4717152b15...64c510ec69d9e63b73ffd5942e802d21a7d14701.

Would it make sense to close this PR? Or maybe redefine this purpose in order to work on the missing tests?

jeremy:
@spastorino indicated that jbuilder should/will get integration tests to cover this 😁

spastorino:
We are going to add integration tests for this during RailsConf :+1:

PR #332, opened 2016-04-06
allow to merge deep hashes

melnikaite:


odedniv:
I think this is better solved in #286 by using `_merge_values` recursively instead of another function (`deep_merge`). Thoughts?

melnikaite:
Agree

PR #333, opened 2016-05-25
Updated changelog to fix a few minor spelling issues

scouttyg:
I was just searching through the changelog for a few things and saw something wasn't being highlighted, and realized it was because of a minor spelling mistake.  I then took the time to do a bit more spelling error cleanup just in case someone else runs into the problem I did.


rwz:
Thanks!

PR #339, opened 2016-06-07
Cleanup: remove relic and confusing 'app' argument

tduek:
The `app` argument isn't used in the `#initializer` block. At a glance, especially because of so much nesting levels, it's confusing where the `app` argument is coming from on line 30 (`app.config.generators`). I didn't notice the other `app` coming in on line 29- thought line 30 was inside the block to `#initializer`.

### DIFF
--- a/lib/jbuilder/railtie.rb
+++ b/lib/jbuilder/railtie.rb
@@ -3,7 +3,7 @@ require 'jbuilder/jbuilder_template'
 
 class Jbuilder
   class Railtie < ::Rails::Railtie
-    initializer :jbuilder do |app|
+    initializer :jbuilder do
       ActiveSupport.on_load :action_view do
         ActionView::Template.register_template_handler :jbuilder, JbuilderHandler
         require 'jbuilder/dependency_tracker'


rwz:
What was wrong with https://github.com/rails/jbuilder/pull/338?

tduek:
@rwz I hadn't read the contribution instructions and #338 wasn't made in a feature branch -_- Didn't know how strict you'd be about that for something this small, but fixed just in case.

rwz:
Looks good, but I'm not sure some versions or rails won't blow up tryng to pass an argument to a block that doesn't take any. Since the variable is not used, maybe it's worth renaming it to something like underscore to make sure block arity stays the same, but at the same time communicating that it's not used. Thgoughts?

tduek:
I like the underscore idea, but the Rails docs show it being used without passing the argument here: [http://api.rubyonrails.org/classes/Rails/Railtie.html](http://api.rubyonrails.org/classes/Rails/Railtie.html) (see Initializers- 2nd heading). Going all the way back to Rails v3.0.0, where Rails::Railtie was introduced, it's described the same way [(http://api.rubyonrails.org/v3.0.0/classes/Rails/Railtie.html)](http://api.rubyonrails.org/v3.0.0/classes/Rails/Railtie.html). 

If you still prefer the underscore approach in light of this, I'm all for it.

tduek:
what are you thinking about this @rwz?

rwz:
👍

rwz:
Thanks

tduek:
You're welcome! Look forward to working with you again!

PR #341, opened 2016-06-13
Fix fragment_cache_key cache invalidation

javan:
Jbuilder currently does not account for an app's [controller-wide `fragment_cache_key`s](https://github.com/rails/rails/blob/master/actionpack/lib/abstract_controller/caching/fragments.rb#L31-L56) when computing its cache keys.

This change adds support, and also aligns Jbuilder's caching setup with [ActionView's cache helpers](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/cache_helper.rb#L219-L238) and [Abstract Controller's fragment caching and instrumentation](https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/actionpack/lib/abstract_controller/caching/fragments.rb#L72-L93).

To test, I made a fresh app with a `fragment_cache_key` call that *should* invalidate the template's cache with every request.
```ruby
# config/environments/development.rb
Rails.application.configure do
  config.x.counter = 0
  config.action_controller.perform_caching = true
  config.cache_store = :mem_cache_store
end

# app/controllers/widgets_controller.rb
class WidgetsController < ApplicationController
  before_action { Rails.application.config.x.counter += 1 }

  fragment_cache_key { Rails.application.config.x.counter }

  def index
  end
end

# app/views/widgets/index.json.jbuilder
json.cache! "unwavering-key" do
  json.config_x_counter Rails.application.config.x.counter
end
```

**Requests using Jbuilder 2.5.0**
```bash
$ curl http://0.0.0.0:3000/widgets.json
{"config_x_counter":1}

$ curl http://0.0.0.0:3000/widgets.json
{"config_x_counter":1} 
```
```log
Started GET "/widgets.json" for 127.0.0.1 at 2016-06-13 18:42:18 -0400
Processing by WidgetsController#index as JSON
  Rendering widgets/index.json.jbuilder
  Rendered widgets/index.json.jbuilder (5.7ms)
Completed 200 OK in 18ms (Views: 10.4ms | ActiveRecord: 0.0ms)

Started GET "/widgets.json" for 127.0.0.1 at 2016-06-13 18:42:20 -0400
Processing by WidgetsController#index as JSON
  Rendering widgets/index.json.jbuilder
  Rendered widgets/index.json.jbuilder (5.7ms)
Completed 200 OK in 10ms (Views: 8.7ms | ActiveRecord: 0.0ms)
```

**Requests using this branch**
```bash
$ curl http://0.0.0.0:3000/widgets.json
{"config_x_counter":1}

$ curl http://0.0.0.0:3000/widgets.json
{"config_x_counter":2}
```
```log
Processing by WidgetsController#index as JSON
  Rendering widgets/index.json.jbuilder
Read fragment jbuilder/views/1/unwavering-key/c9b8ffbaac0ab06acdab1ba36772a5dc (2.7ms)
Write fragment jbuilder/views/1/unwavering-key/c9b8ffbaac0ab06acdab1ba36772a5dc (0.5ms)
  Rendered widgets/index.json.jbuilder (6.5ms)
Completed 200 OK in 19ms (Views: 11.1ms | ActiveRecord: 0.0ms)

Started GET "/widgets.json" for 127.0.0.1 at 2016-06-13 18:39:53 -0400
Processing by WidgetsController#index as JSON
  Rendering widgets/index.json.jbuilder
Read fragment jbuilder/views/2/unwavering-key/c9b8ffbaac0ab06acdab1ba36772a5dc (0.5ms)
Write fragment jbuilder/views/2/unwavering-key/c9b8ffbaac0ab06acdab1ba36772a5dc (0.7ms)
  Rendered widgets/index.json.jbuilder (6.5ms)
Completed 200 OK in 11ms (Views: 9.5ms | ActiveRecord: 0.0ms)
```
Note that the cache was correctly invalidated _and_ the presence logging like you'd see when using `cache` in an .erb template.

/cc @jeremy

### DIFF
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -32,7 +32,7 @@ class JbuilderTemplate < Jbuilder
   #   end
   def cache!(key=nil, options={})
     if @context.controller.perform_caching
-      value = ::Rails.cache.fetch(_cache_key(key, options), options) do
+      value = _cache_fragment_for(key, options) do
         _scope { yield self }
       end
 
@@ -100,6 +100,25 @@ class JbuilderTemplate < Jbuilder
   def _render_partial(options)
     options[:locals].merge! json: self
     @context.render options
+  end
+
+  def _cache_fragment_for(key, options, &block)
+    key = _cache_key(key, options)
+    _read_fragment_cache(key, options) || _write_fragment_cache(key, options, &block)
+  end
+
+  def _read_fragment_cache(key, options = nil)
+    @context.controller.instrument_fragment_cache :read_fragment, key do
+      ::Rails.cache.read(key, options)
+    end
+  end
+
+  def _write_fragment_cache(key, options = nil)
+    @context.controller.instrument_fragment_cache :write_fragment, key do
+      yield.tap do |value|
+        ::Rails.cache.write(key, value, options)
+      end
+    end
   end
 
   def _cache_key(key, options)
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -104,7 +104,13 @@ class JbuilderTemplate < Jbuilder
 
   def _cache_key(key, options)
     key = _fragment_name_with_digest(key, options)
-    key = url_for(key).split('://', 2).last if ::Hash === key
+
+    if @context.respond_to?(:fragment_cache_key)
+      key = @context.fragment_cache_key(key)
+    else
+      key = url_for(key).split('://', 2).last if ::Hash === key
+    end
+
     ::ActiveSupport::Cache.expand_cache_key(key, :jbuilder)
   end
 
--- a/test/jbuilder_template_test.rb
+++ b/test/jbuilder_template_test.rb
@@ -332,6 +332,35 @@ class JbuilderTemplateTest < ActionView::TestCase
     JBUILDER
   end
 
+  test "fragment caching uses fragment_cache_key" do
+    undef_context_methods :fragment_name_with_digest, :cache_fragment_name
+
+    @context.expects(:fragment_cache_key).with("cachekey")
+
+    jbuild <<-JBUILDER
+      json.cache! "cachekey" do
+        json.name "Cache"
+      end
+    JBUILDER
+  end
+
+  test "fragment caching instrumentation" do
+    undef_context_methods :fragment_name_with_digest, :cache_fragment_name
+
+    payloads = {}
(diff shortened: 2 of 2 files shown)


dhh:
👍. It's a bit of a smell that each of the template engines have to implement this logic themselves. Wonder what we can do at a higher level to prevent that.

javan:
Ideally, Jbuilder and others could use the Rails caching helpers more directly, but they don't work with non-string objects. See [action_view/helpers/cache_helper.rb#L227-L238](https://github.com/rails/rails/blob/master/actionview/lib/action_view/helpers/cache_helper.rb#L227-L238)  and [abstract_controller/caching/fragments.rb#L77](https://github.com/rails/rails/blob/master/actionpack/lib/abstract_controller/caching/fragments.rb#L77).

javan:
I think it's worth considering folding Jbuilder into Rails so it can move forward at the same pace. In its current form, it has to remain backwards compatible with a number of Rails versions, and has accumulated gunk like [jbuilder/jbuilder_template.rb#L137-L146](https://github.com/rails/jbuilder/blob/master/lib/jbuilder/jbuilder_template.rb#L137-L146).

dhh:
Yeah, possibly. Currently the default serialization format for Rails API is
on hold until 5.1 to see if something palpable pops up from the
serialization attempts. If not, then default jbuilder and folding jbuilder
in could make sense.

rafaelfranca:
I prefer to create better abstractions in rails han fold all gems to the
same repository. If jbuilder is having problems to evolve in the same pace
be sure that alternative will have more problems to evolve after we fold
jbuilder and stop thinking that alternatives should be possible.

kaspth:
I'll second the notion of finding a better public opt-in API for templates. It's clear that what we do in https://github.com/rails/rails/pull/25825 won't automatically arrive in Jbuilder either, and that's a shame.

From the caching side of things templates just need to be able to read and write fragments (or maybe a fetch fragment method could do). That module should then handle: does the controller enable caching, is there a cache store setup, making sure instrumentation is done correctly, etc. That's the high points as I see it at least.

javan:
> I'll second the notion of finding a better public opt-in API for templates.

For that to happen, the Rails caching helpers will need to work with non-string objects. Jbuilder's "fragments" are objects like Hashes and Arrays. See https://github.com/rails/jbuilder/pull/341#issuecomment-225878084

kaspth:
Ah, right on. Yeah, it's pretty much a non-starter without that support.

Hm, `write_fragment_for` seems pretty attuned to there being a block that evaluates to a string.

bobdutch:
This change removed the ability to use the :force option with cache! now that it uses Rails.cache.read and Rails.cache.write instead of Rails.cache.fetch.

PR #344, opened 2016-07-20
DRY generated view files

sblackstone:
Currently when you generate a scaffold controller, the view files aren't DRY - this patch changes the behavior so that the generated files will look like this:

 app/views/posts/index.json.jbuilder:

```
json.array! @posts, partial: 'posts/post', as: :post
```

app/views/posts/show.json.jbuilder:
```
json.partial! "posts/post", post: @post
```

app/views/posts/_post.json.jbuilder:
```
json.extract! post, :id, :title, :body
json.url post_url(post, format: :json)
```

### DIFF
--- a/lib/generators/rails/jbuilder_generator.rb
+++ b/lib/generators/rails/jbuilder_generator.rb
@@ -20,6 +20,7 @@ module Rails
           filename = filename_with_extensions(view)
           template filename, File.join('app/views', controller_file_path, filename)
         end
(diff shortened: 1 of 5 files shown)


rwz:
Oh, I've just noticed that we've used `attributes_list_with_timestamps` in show action, and `attributes_list` in index... I guess `attributes_list_with_timestamps` everywhere would make more sense in terms of uniformity.

Could you please make the partial use `attributes_list_with_timestamps`?

rwz:
Thanks

PR #352, opened 2016-09-14
Add support to merge! with builder

warmwaffles:
I have been composing builder templates and needed to merge them without a key.

Originally I would have to do 

``` ruby
info = Jbuilder.new do |json|
  json.name("John")
  json.email("john@example.com")
end

combined = Jbuilder.new do |json|
  json.id(1)
  json.merge!(info.attributes!)
end

puts combined.target!
```

Instead I would like to do

``` ruby
info = Jbuilder.new do |json|
  json.name("John")
  json.email("john@example.com")
end

combined = Jbuilder.new do |json|
  json.id(1)
  json.merge!(info)
end

puts combined.target!
```

Subtle change, but it would be nice.

Also added some documentation to the `merge!` method as well.

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -240,8 +240,25 @@ class Jbuilder
     @attributes
   end
 
-  # Merges hash or array into current builder.
-  def merge!(hash_or_array)
+  # Merges hash, array, or any object that responds to the :attributes! method
+  # into current builder.
+  #
+  # Example:
+  #
+  #   info = Jbuilder.new do |json|
+  #     json.name("John")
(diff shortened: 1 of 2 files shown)


warmwaffles:
Also could not get tests running locally, `bundle exec rake test` would just run without output. Super weird.

composerinteralia:
@warmwaffles not totally sure if I got this right, but I ran `appraisal install` and then ran the tests with `appraisal rake test`

warmwaffles:
> not totally sure if I got this right, but I ran appraisal install and then ran the tests with appraisal rake test

I shall give that a shot!

PR #353, opened 2016-09-15
Avoid silently overwriting with merge

composerinteralia:
At the moment `merge!` with an array will overwrite any existing hash
attributes. It should probably raise a `MergeError`. It also makes sense
to raise this error when attempting to merge a hash into an array, or
when trying to merge a non-hash-or-array.

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -270,14 +270,14 @@ class Jbuilder
   def _merge_values(current_value, updates)
     if _blank?(updates)
       current_value
-    elsif _blank?(current_value) || updates.nil?
+    elsif _blank?(current_value) || updates.nil? || current_value.empty? && ::Array === updates
       updates
-    elsif ::Array === updates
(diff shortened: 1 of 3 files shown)


composerinteralia:
@rwz would you like me to rebase on master to keep this up to date?

lib/jbuilder.rb
```diff
@@ -272,15 +272,25 @@ def _merge_values(current_value, updates)
...
     else
-      raise "Can't merge #{updates.inspect} with #{current_value.inspect}"
+      _try_merge(current_value, updates)
     end
   end
 
+  def _try_merge(current_value, updates)
+    if ::Array === updates
+      if ::Array === current_value
+        return current_value + updates
```
rafaelfranca: Can we keep the `if/else` structure and avoid the `return` keywords?

composerinteralia:
@rafaelfranca thanks for your comment! This version is closer to the original.

PR #361, opened 2016-10-08
ignore! function added

abhiQmar:
Use case:

I had to write checks on each line to ignore keys with empty array or empty string values.
```
some_list = []
Jbuilder.encode do |json|
  json.ignore_nil?
  json.string  ''.presence
  json.array   some_list if !some_list.empty?
end
```
the above code now becomes:
```
some_list = []
Jbuilder.encode do |json|
  json.ignore! { |value| value.blank? }
  json.string  ''
  json.array   some_list
end
```


lib/jbuilder.rb
```diff
@@ -106,6 +107,24 @@ def self.key_format(*args)
     @@key_formatter = KeyFormatter.new(*args)
   end
 
+  # If you want to skip adding some values to your JSON hash like empty string
+  # or empty array, you can pass a block.
+  #
+  # Example:
+  #   json.ignore { |value| value.nil? }
```
nateberkopec: Shouldn't these say `ignore!`?

nateberkopec:
@abhiQmar Doesn't jbuilder already check for blank values?

abhiQmar:
@nateberkopec No it doesn't, it makes sense if someone wants to blank arrays, nils and strings(for consistency) in their json output. But if you do not want any blank arrays or empty strings (my use case), you would have to put a check on each and every key-value pair. There is a method(ignore_nil!) to ignore the keys with nil values. I am just trying to extend this functionality so that we can check for anything on our keys cleanly, by sending a block [which if returns true will ignore the key-value pair].

lib/jbuilder.rb
abhiQmar: Yes it should. Thanks for pointing that out.

PR #366, opened 2016-10-21
Update README.md

heycait:
Fix typo - not sure how much giant hash structures like to be 'massaged' :)

Although the description is still not as clear as it could be..

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Jbuilder
 
 Jbuilder gives you a simple DSL for declaring JSON structures that beats
-massaging giant hash structures. This is particularly helpful when the
+messaging giant hash structures. This is particularly helpful when the
 generation process is fraught with conditionals and loops. Here's a simple
 example:
 


rwz:
Thanks!

davidrunger:
I believe that the original wording was correct (albeit maybe slightly unconventional) and that this PR should be reverted.

The sentence in question:

> Jbuilder gives you a simple DSL for declaring JSON structures that beats (massaging|messaging) giant hash structures.

I don't really even know what "messaging giant hash structures" means, but I do know what "massaging giant hash structures means" (the previous wording). Here is a relevant definition of "massage" from http://www.thefreedictionary.com/massaging that applies in this context:

> to manipulate so as to produce a desired result, ex: "to massage data"

That's exactly what Jbuilder does; it manipulates data.

@rwz Let's revert.

davidrunger:
PR to revert: https://github.com/rails/jbuilder/pull/367

heycait:
Hmm I've never heard that usage before. But yes Merriam-Webster lists a similar definition:

"to change (numbers, data, etc.) in a dishonest way in order to deceive people" http://www.merriam-webster.com/dictionary/massage

Either way I think it needs better wording.

davidrunger:
> I think it needs better wording.

@heycait Any suggestions?

heycait:
Or even just say what you said.

Jbuilder gives you a simple DSL for declaring JSON structures that beats manipulating giant hash structures.

davidrunger:
I agree that is better. 👍

davidrunger:
@heycait @rwz  I updated https://github.com/rails/jbuilder/pull/367 to use "manipulating".

PR #370, opened 2016-11-15
Cache root

dhh:
Optimization technique extracted from Basecamp.


woodhull:
Why add this hack instead of accepting the PR that has a more elegant solution for all cases? See #298 

Was there a problem with that PR's approach or anything someone interested in improving jbuilder cache performance generally might be able to contribute?

dhh:
I didn't actually explore #298 in detail prior to this. Will have a look.

PR #372, opened 2016-11-27
Set proper type for jbuilder option to prevent a warning

ojab:
Right now `rails g scaffold_controller` w/ rails-5.0.0.1 produces a warning
>Expected string default value for '--jbuilder'; got true (boolean)


ojab:
…and it's actually Thor change (see erikhuda/thor/pull/435), updated commit message.

PR #375, opened 2016-12-13
Remove a Ruby warning that comes from lib/jbuilder/jbuilder_template.rb

yuki24:
This commit removes the following Ruby warning:

```
lib/jbuilder/jbuilder_template.rb:80: warning: instance variable @cached_root not initialized
```


amatsuda:
I prefer initializing `@cached_root` with nil in `initialize` in this case.

yuki24:
@amatsuda updated.

PR #403, opened 2017-07-03
Remove newly-unsupported Rails versions from the test matrix

georgeclaghorn:
As of eb5d376, 4.2 is the oldest supported version of Rails. Remove 3.x, 4.0, and 4.1 from the test matrix to fix CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,6 @@ rvm:
   - rbx
 
 gemfile:
-  - gemfiles/rails_3_0.gemfile
-  - gemfiles/rails_3_1.gemfile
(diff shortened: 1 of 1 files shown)


georgeclaghorn:
It appears that the newest release of nokogiri dropped support for Ruby versions older than 2.1, causing new failures for Rails 4.2 on Ruby 1.9.3 and 2.0.0. Fixing those in a separate PR.

rafaelfranca:
If we removed from the matrix we also needs to not allow in the gemfile

rafaelfranca:
s/gemfile/gemspec

rafaelfranca:
Ah. David already did

PR #404, opened 2017-07-03
Use a compatible version of Nokogiri to test against Rails 4.2

georgeclaghorn:
Nokogiri 1.7.0 dropped support for versions of Ruby older than 2.1.0, causing CI test failures for Rails 4.2 on Ruby 1.9.3 and 2.0.0. Stick to Nokogiri’s 1.6 release line, which maintains compatibility with Ruby 1.9.3 and newer.


georgeclaghorn:
Failures are for newly-unsupported versions of Rails, removed from the test matrix in #403.

kaspth:
Thanks, @georgeclaghorn!

kaspth:
And the build is green: https://travis-ci.org/rails/jbuilder/builds/249507332 🎉

PR #409, opened 2017-08-27
Switch from MultiJson to the inbuilt JSON library

helaan:
Ruby comes with a JSON library since Ruby 1.9, which makes MultiJson
unneccessary. The usecase mentioned in the README only applies to Rails
4.1 or lower, which is no longer supported by Jbuilder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,21 +243,6 @@ environment.rb for example):
 Jbuilder.key_format camelize: :lower
 ```
 
-Faster JSON backends
---------------------
-
-Jbuilder uses MultiJson, which by default will use the JSON gem. That gem is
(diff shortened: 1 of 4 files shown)


olleolleolle:
@kaspth If I may direct your 👓 attention to this change (really, a fix).

(Has a duplicate PR, see above.)

If you don't have the time, it's totally fine, too.

aelkoussy:
@helaan or anyone who knows this
Out of curiosity, can someone explain why replace MultiJson with default JSON library? why not use OJ or leave MultiJson as it was?
As far as I know OJ is much faster than the standard JSON lib

Thanks

PR #419, opened 2017-10-31
Drop MultiJson dependency

yuki24:
Back in the day, the MultiJson dependency was introduced to add the ability to inject a faster JSON implementation. Time has passed,  and today Jbuilder only has support for Rails 4.2 and older, all major Rubies ship with the `json` gem as stdlib, and Oj 3.0 even provides [Rails mode](https://github.com/ohler55/oj/blob/master/pages/Rails.md#oj-rails-compatibility).

I think it's time to drop the MultiJson dependency and just use `ActiveSupport::JSON::Encoding`. If the user wants to replace the default encoder, they should be responsible for doing so (example for Oj added to the README), and third-party gems should be responsible for ensuring the compatibility with `ActiveSupport::JSON` (which Oj does really well).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+Unreleased
+-----
+
+* Drop the MultiJson dependency
+
 2.6.4
 -----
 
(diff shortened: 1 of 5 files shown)


yuki24:
I jut noticed there's #409, but I'll leave this open since there's a difference between two pull requests (`JSON` vs `ActiveSupport::JSON`).

PR #425, opened 2018-01-26
add :force option to cache!

### LINKED ISSUE #424: cache! force option ignored?
In versions 2.5.0 and older you could do

`json.cache! "key_name", expires_in: 10.minites, force: force? do`
where force? is a method that returns a boolean. If force was true the rails would generate and write the cache. This is broken when I upgrade to 2.6 and greater.

I tried using cache_if! as a workaround but it doesn't expire the cache. Is this a bug or an intentional changed?

bobdutch:
Before change https://github.com/rails/jbuilder/commit/65dfb3305b1c1ebae550c420a6ab6bd0e21271ec you could pass the :force option to cache! This worked because cache! used Rails.cache.fetch which takes the force option. That commit rewrote cache to use Rails.cache.read and Rails.cache.write which don't do anything with the force option. I use :force extensively in my jbuilder template to send a parameter that will clear and write the cache for the next request. I think this is a worthy feature to have even if it was an undocumented and unintentional feature of jbuilder pre-2.6.0
Closes #424

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ Fragment caching is supported, it uses `Rails.cache` and works like caching in
 HTML templates:
 
 ```ruby
-json.cache! ['v1', @person], expires_in: 10.minutes do
+json.cache! ['v1', @person], expires_in: 10.minutes, force: true do
   json.extract! @person, :name, :age
 end
 ```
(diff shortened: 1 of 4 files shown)


HarlemSquirrel:
@pixeltrix it would be awesome if you could spare a moment to review this. Thank you in advance!

MrEmelianenko:
+1 for this

andrewjaanuu:
+1

Issue #378, opened 2017-01-26
Passing object into partial fails if locals: not used


yusufali2205:
Passing objects to partial fails if the below format is used:
`json.comments @post.comments, partial: 'comments/comment', as: :comment, show: false`

It succeeds if locals option is used:
`json.comments @post.comments, partial: 'comments/comment', locals: { show: false }, as: :comment`

SGospodinov:
I tried to reproduce the problem by creating these two tests and adding them to the `test/jbuilder_template_test.rb`. They both expect result records to contain `published` attribute.
```ruby
test "render array of partials with locals passed as a hash" do
  result = jbuild(<<-JBUILDER)
    json.array! BLOG_POST_COLLECTION, partial: "blog_post_with_locals", as: :blog_post, locals: { published: true }
  JBUILDER

  assert_equal result[0]["published"], true
end
```
This test pass.

```ruby
test "render array of partials with locals passed as part of the params" do
  result = jbuild(<<-JBUILDER)
    json.array! BLOG_POST_COLLECTION, partial: "blog_post_with_locals", as: :blog_post, published: true
  JBUILDER

  assert_equal result[0]["published"], true
end
```
But this one raises `ActionView::Template::Error: undefined local variable or method 'published'` error.

I'm using the following partial:

```ruby
BLOG_POST_PARTIAL_WITH_LOCALS = <<-JBUILDER
  json.extract! blog_post, :id, :body
  json.author do
    first_name, last_name = blog_post.author_name.split(nil, 2)
    json.first_name first_name
    json.last_name last_name
  end
  json.published published
JBUILDER
```
Which is a slight modification of one of the existing partials in the test file and also should be added to the `PARTIALS` hash. 
I think the problem is caused by the `_render_partial_with_options` method in `JbuilderTemplate`. If this issue is assessed as a bug, I'll be happy to work on it.

peret:
I also just ran into this after upgrading the gem. This definitely looks like a regression to me. Is there any progress on this issue?

bobdutch:
+1

arcticleo:
+1

eveevans:
`json.array! @post.comments, partial: "comments/comment", as: :comment, locals: { show: false }`

jeremy:
#435

ngouy:
What happend to that?

I still have the issue, and I can't access the linked PR (404)
And can't see any specs related to that under tests/template

(And I have the same issue)

siegfault:
[...]

PR #435, opened 2018-04-17
Fix passing object with partials w/o locals (#378).

siegfault:
This uses @SGospodinov's specs written on the report from @yusufali2205. I ran into this issue as well today and, while I don't think this necessarily the final solution, I wanted to move the conversation along.

Previous behavior: Passing objects to partial fails if the below format is used:
```
json.comments @post.comments, partial: 'comments/comment', as: :comment, show: false
```

It succeeds if locals option is used:
```
json.comments @post.comments, partial: 'comments/comment', locals: { show: false }, as: :comment
```

Swapping to options is a change to revitalize the conversation. I'm not
sure it's a permanent solution.

### DIFF
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -104,7 +104,7 @@ class JbuilderTemplate < Jbuilder
   private
 
   def _render_partial_with_options(options)
-    options.reverse_merge! locals: {}
+    options.reverse_merge! locals: options.except(:partial, :as, :collection)
     options.reverse_merge! ::JbuilderTemplate.template_lookup_options
     as = options[:as]
 
@@ -199,7 +199,6 @@ class JbuilderTemplate < Jbuilder
     when ::Hash
       # partial! partial: 'name', foo: 'bar'
       options = name_or_options
-      options[:locals] ||= options.except(:partial, :as, :collection)
     else
       # partial! 'name', locals: {foo: 'bar'}
       if locals.one? && (locals.keys.first == :locals)
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -111,8 +111,9 @@ class JbuilderTemplate < Jbuilder
     if as && options.key?(:collection)
       as = as.to_sym
       collection = options.delete(:collection)
+      locals = options.delete(:locals) || {}
       array! collection do |member|
-        member_locals = options.clone
+        member_locals = options.clone.merge(locals)
         member_locals.merge! collection: collection
         member_locals.merge! as => member
         _render_partial options.merge(locals: member_locals)
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -111,9 +111,9 @@ class JbuilderTemplate < Jbuilder
     if as && options.key?(:collection)
       as = as.to_sym
       collection = options.delete(:collection)
-      locals = options.delete(:locals) || {}
+      locals = options.delete(:locals)
       array! collection do |member|
-        member_locals = options.clone.merge(locals)
+        member_locals = locals.clone
         member_locals.merge! collection: collection
         member_locals.merge! as => member
         _render_partial options.merge(locals: member_locals)
@@ -199,6 +199,7 @@ class JbuilderTemplate < Jbuilder
     when ::Hash
       # partial! partial: 'name', foo: 'bar'
       options = name_or_options
+      options[:locals] ||= options.except(:partial, :as, :collection)
     else
       # partial! 'name', locals: {foo: 'bar'}
       if locals.one? && (locals.keys.first == :locals)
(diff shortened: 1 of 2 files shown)


siegfault:
@pixeltrix would love any thoughts or advice on this one that you can provide! Thanks!

lib/jbuilder/jbuilder_template.rb
```diff
@@ -111,9 +111,8 @@ def _render_partial_with_options(options)
     if as && options.key?(:collection)
       as = as.to_sym
       collection = options.delete(:collection)
-      locals = options.delete(:locals)
       array! collection do |member|
-        member_locals = locals.clone
+        member_locals = options.clone
```
yusufali2205: Does it mean using locals won't work now?
```
json.comments @post.comments, partial: 'comments/comment', locals: { show: false }, as: :comment
```

siegfault: I wrote a spec that I'm pretty sure confirms your suspicion that locals don't work now but haven't pushed it. I've been pretty swamped the last couple weeks but should be much freer next week to take a more in depth look. Thanks for taking a look @yusufali2205!

siegfault:
@yusufali2205 just pushed a change that highlights and seems to fix the concern you had. I don't think it's a good solution for the reasons I mentioned in the commit that I'll copy here. Definitely looking for advice!

> This is not a good solution because it puts all of options, related to
locals or not, into the locals section of options. Will want advice for
a better solution, but this does make the spec pass.

test/jbuilder_template_test.rb
```diff
@@ -123,23 +123,39 @@ def assert_collection_rendered(result, context = nil)
...
     result = jbuild(<<-JBUILDER)
-      json.partial! "partial"
+      json.partial! "partial", foo: 'hello'
     JBUILDER
 
     assert_equal "hello", result["content"]
   end
 
-  test "partial! + locals via :locals option" do
+  test "partial! + locals without :partial key with :locals key" do
```
siegfault: Wouldn't mind some help making these descriptions less bulky.

test/jbuilder_template_test.rb
```diff
@@ -123,23 +123,39 @@ def assert_collection_rendered(result, context = nil)
...
+
+  test "partial! + locals without :partial key without :locals key" do
+    result = jbuild(<<-JBUILDER)
+      json.partial! "partial", foo: "howdy"
+    JBUILDER
+
+    assert_equal "howdy", result["content"]
+  end
+
+  test "partial! + locals with :partial key without :locals key" do
```
siegfault: This is the spec that currently would fail with the above 
```diff
-  "_partial.json.jbuilder"  => "foo ||= 'hello'; json.content foo",
+  "_partial.json.jbuilder"  => "json.content foo",
```
change.

test/jbuilder_template_test.rb
```diff
@@ -44,7 +44,7 @@ def initialize(id, name)
 ActionView::Template.register_template_handler :jbuilder, JbuilderHandler
 
 PARTIALS = {
-  "_partial.json.jbuilder"  => "foo ||= 'hello'; json.content foo",
```
siegfault: Removed to be able to reproduce the undefined method error. This caused me to add
```ruby
-      json.partial! "partial"
+      json.partial! "partial", foo: 'hello'
```
Happy to discuss if we'd rather go back to defining its own partial.

lib/jbuilder/jbuilder_template.rb
```diff
@@ -199,6 +199,7 @@ def _render_explicit_partial(name_or_options, locals = {})
     when ::Hash
       # partial! partial: 'name', foo: 'bar'
       options = name_or_options
+      options[:locals] ||= options.except(:partial, :as, :collection)
```
siegfault: If we want to take these out of the original options hash, I think we'd need to do something like
```ruby
options[:locals] ||= options.except(:partial, :as, :collection).keys.each_with_object({}) do |local, hash|
  hash[local] = options.delete(local)
end
```
Thoughts?

siegfault:
I think this is now in a better place and ready for review @yusufali2205 / @pixeltrix. Happy to make any changes and would love any feedback. Thanks!

siegfault:
@pixeltrix let me know if you have any advice or if you want me to ping someone else. Thanks!

siegfault:
@rwz pinging you since it looks like you were getting assigned PRs before @pixeltrix. Any ideas on who I should pull in to take alook at this? Thanks!

@yusufali2205 also pinging you since you originally reported the issue and reviewed this earlier. This seems to fix the issue for me. Mind making sure that it also does for you? Thanks!

siegfault:
Hi @rafaelfranca! Pinging you since you seem to be getting assigned on PRs now (though it's noting that the repo may be misconfigured). Are you the right person to take a look? If not, do you know who would be? Either way, thanks!

siegfault:
Hi @dhh! Pinging you since you merged a PR in the repo last week. Not necessarily asking for review unless you want to take a look, but since the repo is misconfigured and I haven't gotten much traction here, was wondering if you had suggestions for who to reach out to. Thanks!

dhh:
Hey, sorry you haven't gotten any traction on this yet. Thank for you for keep pushing! 

This looks correct to me. Merging. Thanks again!

siegfault:
Thanks!

Issue #330, opened 2016-03-28
nil! after cache! not working reliable


ThomasPr:
I came up with a really strange issue tonight: The following snippet produces an error on the THRID request. The first and second requests result in the expected json.

```ruby
json.array! [1, 2, 3] do |index|
  json.cache! index do
    json.key1 "anything"
  end
  json.key2 do
    json.nil!
  end
end
```

```
ActionView::Template::Error (Failed to add "key2" property to null object):
    2:   json.cache! index do
    3:     json.key1 "anything"
    4:   end
    5:   json.key2 do
    6:     json.nil!
    7:   end
    8: end
  app/views/example/example.json.jbuilder:5:in `block in _app_views_example_example_json_jbuilder___2645842209584969737_70123296568040'
  app/views/example/example.json.jbuilder:1:in `_app_views_example_example_json_jbuilder___2645842209584969737_70123296568040'
```

I created a [minimal example](https://github.com/ThomasPr/example). Please execute `curl -X "GET" "http://localhost:3000"` three times to get the error.

```
activesupport (4.2.6) lib/active_support/proxy_object.rb:10:in `raise'
jbuilder (2.4.1) lib/jbuilder.rb:265:in `_merge_block'
jbuilder (2.4.1) lib/jbuilder.rb:38:in `set!'
jbuilder (2.4.1) lib/jbuilder/jbuilder_template.rb:74:in `set!'
jbuilder (2.4.1) lib/jbuilder.rb:66:in `method_missing'
app/views/example/example.json.jbuilder:5:in `block in _app_views_example_example_json_jbuilder___2645842209584969737_70123296568040'
jbuilder (2.4.1) lib/jbuilder.rb:298:in `block (2 levels) in _map_collection'
jbuilder (2.4.1) lib/jbuilder.rb:305:in `_scope'
jbuilder (2.4.1) lib/jbuilder.rb:298:in `block in _map_collection'
jbuilder (2.4.1) lib/jbuilder.rb:297:in `map'
jbuilder (2.4.1) lib/jbuilder.rb:297:in `_map_collection'
jbuilder (2.4.1) lib/jbuilder.rb:188:in `array!'
jbuilder (2.4.1) lib/jbuilder/jbuilder_template.rb:64:in `array!'
app/views/example/example.json.jbuilder:1:in `_app_views_example_example_json_jbuilder___2645842209584969737_70123296568040'
actionview (4.2.6) lib/action_view/template.rb:145:in `block in render'
```

[...]

PR #436, opened 2018-05-13
fix cache bug

BenZhang:
Hi,
I ran into a very strange situation.
```ruby
json.array! [1,2,3] do |i|
  json.cache! 'test' do
    json.name 'cache'
  end
  json.id i
end
```
The code above will print 
```javascript
[
  {"name":"cache","id":1},
  {"name":"cache","id":3},
  {"name":"cache","id":3}
]
```
If we move json.id on the top, it will work. The reason for that is when use cache! in the beginning of the block it will retrieve the same cached object from previous iteration and overwrite it. I think it is the same reason for this issue #330

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -295,8 +295,7 @@ class Jbuilder
 
   def _map_collection(collection)
     collection.map do |element|
-      element_result = _scope{ yield element }
-      element_result == BLANK ? BLANK : element_result.clone
+      _scope{ yield element }
     end - [BLANK]
   end
 
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -295,7 +295,8 @@ class Jbuilder
 
   def _map_collection(collection)
     collection.map do |element|
-      _scope{ yield element }
+      element_result = _scope{ yield element }
+      element_result == BLANK ? BLANK : element_result.clone
     end - [BLANK]
   end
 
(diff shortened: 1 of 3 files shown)


BenZhang:
After further investigation, this will only happen when using memory_store for the cache. It is actually caused by the behaviour of Rails cache. Please see below example

```ruby
Rails.cache.write('user', {id: 1})
a = Rails.cache.read('user')
#=> {:id=>1}
a[:name] = "Hello"
Rails.cache.read('user')
#=> {:id=>1, :name => "Hello"}
```

ryanb:
I ran into this today. It would be great to get it fixed! Can someone address this PR?

EvTi:
One more bug, i think

```
json.cache_collection! [1,2,3] do |category|
  next
end
```

result 

```
{}
{}
{}
```

dhh:
I know this is a million years old, but just stumbled over it. It strikes me as odd this is only happening in the memory store. That implies to me it should be fixed in the memory store. Anyone interested in digging deeper?

pixeltrix:
@dhh what version was this in? I've tried using @BenZhang's test and his cache commands and can't seem to reproduce in any version of 7.x

Looking at his sequence of commands suggests it's an example of rails/rails#36956:

``` ruby
Rails.cache.write('user', {id: 1})
a = Rails.cache.read('user')
#=> {:id=>1}
a[:name] = "Hello"
Rails.cache.read('user')
#=> {:id=>1, :name => "Hello"}
```
that's not been applicable since rails/rails#37210.

dhh:
Ha. Makes sense. The initial report was ancient but hadn't seen any updates. Will close. Thanks @pixeltrix 🙏

PR #440, opened 2018-06-20
Make extract! accept objects with internal hash implementation

bacchir:
In a recent project I've experienced an odd situation of having an object that implements some `Hash` behavior and, even though it has methods such as `[]` and `fetch`, `jbuilder.extract!` did not work with it. So I just created this little PR with some tests to make `jbuilder` more agnostic.

Here's an example (also used in the tests) of the object I'm trying to pass to jbuilder:

```
class PersonWithHash
  attr_reader :name, :collection

  def initialize(name, age)
    @collection = { age: age }
    @name = name
  end

  delegate :[], :fetch, to: :@collection
end
```

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -213,10 +213,8 @@ class Jbuilder
   #
   #   json.(@person, :name, :age)
   def extract!(object, *attributes)
-    if ::Hash === object
-      _extract_hash_values(object, attributes)
-    else
-      _extract_method_values(object, attributes)
+    attributes.each do |key|
+      _set_value key, _extract_value(object, key)
     end
   end
 
@@ -252,12 +250,8 @@ class Jbuilder
 
   private
 
-  def _extract_hash_values(object, attributes)
-    attributes.each{ |key| _set_value key, object.fetch(key) }
-  end
-
-  def _extract_method_values(object, attributes)
-    attributes.each{ |key| _set_value key, object.public_send(key) }
+  def _extract_value(object, attribute)
+    object.respond_to?(attribute) ? object.public_send(attribute) : object.fetch(attribute)
   end
 
   def _merge_block(key)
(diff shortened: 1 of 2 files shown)


lib/jbuilder.rb
```diff
@@ -252,12 +250,8 @@ def target!
...
   private
 
-  def _extract_hash_values(object, attributes)
-    attributes.each{ |key| _set_value key, object.fetch(key) }
-  end
-
-  def _extract_method_values(object, attributes)
-    attributes.each{ |key| _set_value key, object.public_send(key) }
+  def _extract_value(object, attribute)
+    object.respond_to?(attribute) ? object.public_send(attribute) : object.fetch(attribute)
```
robin850: The risk is pretty low but still, it should rather be:

~~~ruby
object.respond_to?(:fetch) ? object.fetch(attribute) : object.public_send(attribute)
~~~

Otherwise, if a hash or hash-like object has a `:object_id`, `:hash` or any of its method as a key, the wrong value may be returned.

Could you fix this and add a test to ensure that a `Hash` with such key would be properly handled please ?

bacchir: Sorry @robin850 I just saw your comment now. Sure I'll change it 😊

bacchir:
@robin850 I changed the implementation the way you suggested, however I changed `respond_to?(:fetch)` to `respond_to?(:[])` to avoid problems when an object have a custom `fetch` method for other purposes.

lib/jbuilder.rb
bacchir: @robin850 I changed the implementation the way you suggested, however I changed respond_to?(:fetch) to respond_to?(:[]) to avoid problems when an object have a custom fetch method for other purposes.

bacchir:
@robin850 I changed the implementation the way you suggested, however I changed respond_to?(:fetch) to respond_to?(:[]) to avoid problems when an object have a custom fetch method for other purposes.

PR #448, opened 2018-10-29
Remove timestamps from generated partial

### LINKED ISSUE #437: rails g scaffold --no-timestamps not works
rails version: 5.2.0
jbuilder version: 2.7.0

The generated app/views/model/_model.json.jbuilder still contains created_at & updated_at, even with --no-timestamps

robin850:
Hello,

This pull request fixes #437.

When the `--no-timestamps` option is passed to the scaffold generator, Active Record will properly remove the `created_at` and `updated_at` columns from the migration file but the generated Jbuilder partial will still reference these fields.

Have a nice day !

### DIFF
--- a/lib/generators/rails/jbuilder_generator.rb
+++ b/lib/generators/rails/jbuilder_generator.rb
@@ -10,6 +10,8 @@ module Rails
 
       argument :attributes, type: :array, default: [], banner: 'field:type field:type'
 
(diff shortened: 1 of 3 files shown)


olleolleolle:
@robin850 Oh, this looks super-clear - and fixes the issue.

PR #449, opened 2018-10-30
Tiny cleanup in the tests

robin850:
Hello,

That's just a tiny pull request that cleans up a bit the test files.

Use `assert_nil` rather than `assert_equal nil, foo` since this behavior will be change to raise an error in the upcoming major version of Minitest.

Also, remove the various checks to support older versions of Rails. We are not supporting any version older than 4.2 since eb5d376. Also, `ActiveSupport.test_order` has been added in Rails 4.2 (c.f. https://github.com/rails/rails/commit/53e877f7d9291b2bf0b8c425f9e32ef35829f35b).

Have a nice day !

### DIFF
--- a/test/jbuilder_test.rb
+++ b/test/jbuilder_test.rb
@@ -75,7 +75,7 @@ class JbuilderTest < ActiveSupport::TestCase
     end
 
     assert result.has_key?('content')
-    assert_equal nil, result['content']
+    assert_nil result['content']
   end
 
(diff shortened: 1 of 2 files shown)


olleolleolle:
@kaspth 👋 Look, this is a neat change!

kaspth:
@olleolleolle you got me! 😄

@robin850 hey Robin, thanks for doing this!

PR #453, opened 2019-02-04
Fix deprecation warning for JBuilder templates

### LINKED ISSUE #452: DEPRECATION WARNING in new --edge app
```
$ ruby -v
ruby 2.6.1p33 (2019-01-30 revision 66950) [x86_64-linux]
$ rails -v
Rails 6.0.0.beta1
$ rails new blog --edge > /dev/null 2>&1
$ cat blog/log/development.log 
DEPRECATION WARNING: Single arity template handlers are deprecated.  Template handlers must
now accept two parameters, the view object and the source for the view object.
Change:
  >> Class#call(template)
To:
  >> Class#call(template, source)
 (called from block (2 levels) in <class:Railtie> at /home/u/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/jbuilder-2.8.0/lib/jbuilder/railtie.rb:8)
```

tenderlove:
This commit brings the JBuilder template handler up to work with Rails 6
and no deprecation warnings.  The `JbuilderHandler.call` method should
work with older Rails versions too.

Fixes #452

### DIFF
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -224,9 +224,10 @@ class JbuilderHandler
   cattr_accessor :default_format
   self.default_format = Mime[:json]
 
-  def self.call(template)
+  def self.call(template, source = nil)
+    source ||= template.source
     # this juggling is required to keep line numbers right in the error
-    %{__already_defined = defined?(json); json||=JbuilderTemplate.new(self); #{template.source}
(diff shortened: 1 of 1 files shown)


jasnow:
coffee-rails needs this too.

morgoth:
Can we have a jbuilder release with this fix?

phoet:
`gem 'jbuilder', github: 'rails/jbuilder', branch: 'master'`

would be great to have a gem release though!

PR #459, opened 2019-04-05
Add rich_text, attachment, and attachments fields support

excid3:
I just added support for generating models with `rich_text`, `attachment` and `attachments` fields for ActionText and ActiveStorage to Rails 6 recently.

This updates jbuilder's scaffold generator to also support these fields and retains the old functionality for older Rails versions.

### DIFF
--- a/lib/generators/rails/jbuilder_generator.rb
+++ b/lib/generators/rails/jbuilder_generator.rb
@@ -50,6 +50,10 @@ module Rails
 
           attributes.map { |a| ":#{a}"} * ', '
         end
+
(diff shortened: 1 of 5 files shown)


excid3:
@rafaelfranca wanted to bump this. Would you consider merging this one since Rails 6 already has these generator fields in https://github.com/rails/rails/pull/35805? 

Jbuilder overrides the generators so they would appear to not work in a new Rails 6 app until jbuilder is also updated. Thanks!

dpaola2:
Voicing my support for this functionality

PR #468, opened 2019-10-29
Fix warnings on Ruby 2.7

r7kamura:
> warning: Capturing the given block using Proc.new is deprecated; use `&block` instead

This patch converts `Proc.new` calls into `&block`.


assembler:
When will this be released as new gem version?

PR #471, opened 2019-11-22
namespaced_file_path was removed from railties

kdiogenes:


olleolleolle:
This follows Rails' change, here's a direct link to how Rails did when they switched away from the removed method:

https://github.com/rails/rails/commit/595fa6e9225a7a33806efa74fcbe54decdc37510

PR #476, opened 2020-03-03
Use unless instead of if not

liroyleshed:


rafaelfranca:
Thank you for the pull request but we prefer `if !` over `unless/else`

PR #477, opened 2020-03-16
README: Add a TravisCI build status badge

olleolleolle:
This points at the master branch.


rafaelfranca:
Thank you for the pull request but we moved all badges to the contributing file since they are more related to contributors than users. https://github.com/rails/jbuilder/blob/master/CONTRIBUTING.md#contributing-to-jbuilder

PR #481, opened 2020-03-17
CI: Set dist: trusty for Rubinius

olleolleolle:
This PR

- configures the Travis "dist" to a Rubinius-supporting one.
- adds gems that rbx needs to the gemspec

When we know more which of these matrix combination elements are even theoretically supported, we may be able to make the matrix smaller.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ rvm:
   - 2.5.5
   - 2.6.2
   - ruby-head
-  - rbx-3.107
 
 gemfile:
   - gemfiles/rails_5_0.gemfile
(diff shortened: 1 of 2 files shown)


olleolleolle:
Racc/parser.rb missing seems hard to do anything about. 

The depreciation warning from mocha is fixable but changes nothing for the CI situation for Rubinius. 

Perhaps it's not possible to run Rubinius with mocha in the test suite

olleolleolle:
https://github.com/teamcapybara/capybara/pull/1194/files#r7912212 This comment in Capybara (2013) explains.

olleolleolle:
@kaspth Thanks for the confidence, but we're closer but not quite there yet.

The `json` extension can not be built completely, yet, on Rubinius.

> Dear Rubinius team & community: I made a tiny PR trying to get to install json support. What's the best way for me to do that? rails/jbuilder#481
> my current failure is: JSON extension fails on ST_CONTINUE not being set up. I learned that in the oniguruma build/ directory in the platform's repository, there was [a st.h header file](https://github.com/rubinius/rubinius/blob/75086f2b2cc92302b54176db0250ec6635adfcc8/build/libraries/oniguruma/st.h) which defined it as part of an enum. https://travis-ci.org/github/rails/jbuilder/jobs/663411968#L1087 Travis link to my failure.
> https://github.com/rubinius/rubinius/issues/3795#issuecomment-570830492 I read this comment, and now I thought "perhaps I'm trying to do things that... won't work".
>

kaspth:
Ah, gotcha. Figured the 5.2 fix would carry over here, in any case this is better than what was before even if not all the way there. Thanks for taking this on 🙏

rafaelfranca:
Can we remove rubinius from our build matrix? It is not a "Ruby" anymore.

PR #488, opened 2020-06-15
Test with Ruby 2.7

aried3r:
Also reduce need to micro-manage patch versions of Ruby versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,12 @@ before_install:
   - "gem install bundler -v '<2'"
 
 rvm:
-  - 2.2.10
-  - 2.3.8
-  - 2.4.6
-  - 2.5.5
-  - 2.6.2
+  - 2.2
+  - 2.3
+  - 2.4
+  - 2.5
+  - 2.6
+  - 2.7
   - ruby-head
 
 gemfile:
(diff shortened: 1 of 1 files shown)


aried3r:
I tried a few things, but Ruby 2.2 makes problems. I'm not sure if dropping Ruby 2.2 is an option or we should go the extra mile. Either with different builds stages, jobs or something [like suggested here](https://travis-ci.community/t/suggestion-for-how-to-update-rubygems-bundler/1681).

Friendly ping @kaspth, @olleolleolle.

olleolleolle:
@kaspth Do you have an take on the support schedule for this project? Which Ruby versions are appropriate in the matrix?

olleolleolle:
The issue in the build, failing 2.2 is:

```
$ gem install bundler

ERROR:  Error installing bundler:
	bundler requires Ruby version >= 2.3.0.
The command "gem install bundler" failed and exited with 1 during .
```

olleolleolle:
Ah, we have had movement in the project, and this PR has been superseded!

https://github.com/rails/jbuilder/blob/master/.travis.yml#L14 lists 2.7.1 in the matrix.

I think this means that we got what we came for, and that this Pull Request can be closed as done!

PR #494, opened 2020-12-24
Silently attempt to fetch non-existent hash value

timfjord:
When the `array!` method attempts to read a non-existent hash value it raises the `KeyError` error because it uses `Hash#fetch` method under the hood without the second argument.
This case is not covered with tests(the Hash case is not covered at all) so I assume it is not the expected behavior.
I cannot find any arguments when this behavior might be useful.

This PR fixes this issue and makes sure an attempt to read a non-existent hash value returns `nil`

### DIFF
--- a/lib/jbuilder.rb
+++ b/lib/jbuilder.rb
@@ -254,7 +254,7 @@ class Jbuilder
   private
 
   def _extract_hash_values(object, attributes)
-    attributes.each{ |key| _set_value key, object.fetch(key) }
+    attributes.each{ |key| _set_value key, object.fetch(key, BLANK) }
   end
 
   def _extract_method_values(object, attributes)
--- a/test/jbuilder_test.rb
+++ b/test/jbuilder_test.rb
@@ -450,6 +450,34 @@ class JbuilderTest < ActiveSupport::TestCase
     assert_equal       2, result.second['id']
   end
 
+  test 'extract hash keys directly from array' do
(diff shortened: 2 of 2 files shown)


dhh:
Seems like its better to raise on a missing attribute than to silently return nil? Seems like a way you'd get screwed on spelling errors?

timfjord:
I'd consider that this piece of code
```ruby
data = [{ id: 1, content: 'hello', meta: 'meta' }, { id: 2, content: 'world' }] 

json.array! data, :id, :content, :meta
```
is a shortcut for
```ruby
data = [{ id: 1, content: 'hello', meta: 'meta' }, { id: 2, content: 'world' }] 

json.array! data do |hash|
  json.id hash[:id]
  json.content hash[:content]
  json.meta hash[:meta]
end
```
So I would expect them to behave similarly.

Of course we can say, that instead of `Hash#[]` we should have used `Hash#fetch`, plus, indeed, `#fetch` can help us with 
spelling errors, so maybe we could handle that with an additional option? 
```ruby
json.array! data, :id, :content, :meta, fetch: true # assuming that `Hash#[]` is the default method
```
or 
```ruby
json.array! data, :id, :content, :meta, silent: true # assuming that `Hash#fetch` is the default method
```

timfjord:
Hm, we could even handle that completely differently and intrude `default` option so it can be used for both hashes and objects
```ruby
json.array! data, :id, :content, :meta, default: nil
```

PR #499, opened 2021-02-10
Rails use 'main' branch now.

voxik:
Opening this PR, because test suite fails testing against Rails 6.1 and nobody can notice, because it is not properly executed due to Rails branch name changes.


voxik:
> because test suite fails testing against Rails 6.1 and nobody can notice, because it is not properly executed due to Rails branch name changes.

Actually, this might also be Ruby 3.0 thing ...

PR #502, opened 2021-02-22
Update travis.yml to reflect the current state of maintained Ruby/Rails versions

yuki24:
I noticed the `travis.yml` is a bit out of date and doesn't include the latest versions of Ruby and Rails. I have also fixed the builds for Rails 6.1.


test/jbuilder_template_test.rb
```diff
@@ -290,7 +290,7 @@ def render(*args)
 
     def render_without_parsing(source, assigns = {})
       view = build_view(fixtures: PARTIALS.merge("source.json.jbuilder" => source), assigns: assigns)
-      view.render(template: "source.json.jbuilder")
+      view.render(template: "source")
```
yuki24, line 293: This fixes the builds for Rails 6.1 and later.

PR #504, opened 2021-06-25
Remove reliance on ERBTracker from rails

HParker:
Use dependency tracker that does not rely on a rails dependency tracker.
This prevents problems if rails chooses to remove or change ERBTracker.

Related to: https://github.com/rails/rails/pull/42458


lib/jbuilder/dependency_tracker.rb
```diff
@@ -54,8 +55,5 @@ def indirect_dependencies
       end
     end
   end
-
-  ::Jbuilder::DependencyTracker = Class.new(dependency_tracker::ERBTracker)
-  ::Jbuilder::DependencyTracker.send :include, ::Jbuilder::DependencyTrackerMethods
```
rafaelfranca, line 59: Should we delete `DependencyTrackerMethods`?

HParker:
@rafaelfranca Sorry for the huge delay. We should remove `DependencyTrackerMethods`. Updated, but kept the tests testing the new method of resolving templates.

PR #526, opened 2021-12-23
Run CI against Rails 7.0

okuramasafumi:


petergoldstein:
Would it make sense to get a combined PR of this and #528 ?  It seems odd that JBuilder is not running in CI against both Ruby 3.1 and Rails 7.0.x

cc: @berkos

berkos:
> Would it make sense to get a combined PR of this and #528 ? It seems odd that JBuilder is not running in CI against both Ruby 3.1 and Rails 7.0.x
> 
> cc: @berkos

Makes sense. I am happy to cherry-pick this commit in my PR if we decide so.

.github/workflows/ruby.yml
```diff
@@ -69,6 +70,16 @@ jobs:
...
+          - ruby: 2.2
+            gemfile: rails_7_0
+          - ruby: 2.3
+            gemfile: rails_7_0
+          - ruby: 2.4
+            gemfile: rails_7_0
+          - ruby: 2.5
+            gemfile: rails_7_0
+          - ruby: 2.6
+            gemfile: rails_7_0
```
dhh, line 82: We shouldn't be testing on Ruby versions that are not supported by Rails 7. Rails 7 only supports 2.7+

berkos: 👋 I believe this is the exclusion list in the `.yml` so it won't run on those versions. 

Given on the discussion [here](https://github.com/rails/jbuilder/pull/526#issuecomment-1028934958) and that this PR has not been updated recently I've included Rails 7 and Ruby 3.1 in a single PR [here](https://github.com/rails/jbuilder/pull/528) that originally had only Ruby 3.1 testing support.

PR #538, opened 2022-08-25
Add missing ':see_other' status code in generated destroy controller method

ldeld:
See https://github.com/rails/rails/pull/45383

When scaffolding a model (with `rails g scaffold ModelName`), the destroy method in the generated controller is missing a 302 status code. Adding this status code is recommended in the [Getting Started guide](https://guides.rubyonrails.org/getting_started.html#deleting-an-article), so it makes sense to add it to the generated controller as well (especially since the not having the status code can cause issues when using a turbo link with `data-turbo-method='delete'`)


ldeld:
Tests failed for rails_head, but I don't believe those failures are related to this PR (the errors also happen when running tests on from the main branch)

PR #552, opened 2023-08-28
Fix require path of dependency_tracker in railtie.rb

jalyna:
```
LoadError: cannot load such file -- jbuilder/dependency_tracker
```

Introduced in https://github.com/rails/jbuilder/commit/710979958aa012f4ff21800abf5bd50b717df0a6


amatsuda:
Merging this since I could reproduce the situation on CI via #556. Thanks @jalyna!

PR #556, opened 2024-01-03
Run Railtie from the tests so that we can detect bugs inside Railtie

amatsuda:
#552 seems like a valid fix. The `require` statement seems to be requiring something that no longer exists. Then why isn't the CI failing? It should be because the Jbuilder Railtie is not executed during the tests.

So, here's an attempt to make the CI run the Railtie. Let's see if the CI properly fails now...

### DIFF
--- a/lib/jbuilder/railtie.rb
+++ b/lib/jbuilder/railtie.rb
@@ -16,9 +16,11 @@ class Jbuilder
           end
         end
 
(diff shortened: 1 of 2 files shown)


amatsuda:
Confirmed that tests were failing with this patch https://github.com/rails/jbuilder/actions/runs/7401605179, then merged #552 and rebased, and all the tests are passing now 🎉

PR #557, opened 2024-01-24
Support Strict Locals

seanpdoyle:
When scaffolding out partials, generate template files with a [Strict Locals][] magic comment based on the name of the resource (for example, `# locals: (post:)` when generating `_post.json.jbuilder`).

[Strict Locals]: https://guides.rubyonrails.org/action_view_overview.html#strict-locals

### DIFF
--- a/lib/generators/rails/templates/partial.json.jbuilder
+++ b/lib/generators/rails/templates/partial.json.jbuilder
@@ -1,3 +1,5 @@
+# locals: (<%= singular_table_name %>:)
+
 json.extract! <%= singular_table_name %>, <%= full_attributes_list %>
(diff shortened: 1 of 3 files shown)


seanpdoyle:
This is facing issues similar to https://github.com/rails/rails/issues/49761 (which was resolved by https://github.com/rails/rails/pull/49782).

seanpdoyle:
@casperisfine I've shared some thoughts in https://github.com/rails/rails/pull/50864#pullrequestreview-1843455961, but it seems like the public API around strict and implicit locals is insufficient for templating engines.

dhh:
I'm not a fan of strict locals, and I'm not interested in them being used by default.

PR #569, opened 2024-09-01
Redirect to `@record` or path in controller generator

jeromedalbert:
Partially addresses https://github.com/rails/rails/issues/52756

### Problem

There is some discrepancy between Rails controller generators and jbuilder controller generators when it comes to redirects.

### Solution

This PR changes the controller generator to generate redirects with `redirect_to @post` or `_path`, in order to match the Rails codebase generators as close as possible.

### DIFF
--- a/lib/generators/rails/templates/controller.rb
+++ b/lib/generators/rails/templates/controller.rb
@@ -30,7 +30,7 @@ class <%= controller_class_name %>Controller < ApplicationController
 
     respond_to do |format|
       if @<%= orm_instance.save %>
-        format.html { redirect_to <%= show_helper %>, notice: <%= %("#{human_name} was successfully created.") %> }
+        format.html { redirect_to <%= redirect_resource_name %>, notice: <%= %("#{human_name} was successfully created.") %> }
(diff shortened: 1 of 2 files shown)


jeromedalbert:
Tests are failing for Rails 5 as `singular_route_name` is not available in that version. I am wondering if a better fix would be to stop testing Ruby version 2.7 and Rails 5.

skipkayhil:
Would `redirect_to @post` work in this case? I think that's what the Rails' templates are generating now (however, `redirect_to @post` is still redirecting with a full URL which I think DHH is wanting to change)

dhh:
Yes, let's change to `redirect_to @post` so it matches the generator in Rails itself. They should be as close as possible.

jeromedalbert:
PR updated. Rails 5 tests are failing for Rails 5 and 5.1 because [`redirect_resource_name`](url) was [added](https://github.com/rails/rails/pull/30729) in Rails 5.1.

I would suggest not testing against these end of life versions. I can edit this PR to fix this or make a separate one.

PR #573, opened 2024-09-14
Use new `params.expect` syntax instead of `params.require`

jeromedalbert:
## Problem

PR https://github.com/rails/rails/pull/51674 updated controller generator templates in the Rails codebase. But, as DHH [explained](https://github.com/rails/rails/pull/51674#issuecomment-2336471213), we need to do the same update to templates in jbuilder, because they are used by default for new Rails apps.

## Solution

Port the changes from https://github.com/rails/rails/pull/51674 to this codebase, only for Rails versions 8.0+:

- Change templates to use `params.expect(post: [ :title, ... ])` instead of `params.require(:post).permit(:title, ...)`.

  Notice the spaces around `[` and `]` in order to be compliant with `rubocop-rails-omakase` rules, which are enabled by default for new Rails apps.
- Change templates to use `@post = params.expect(:id)` instead of `@post = params[:id]`.

cc @martinemde

### DIFF
--- a/lib/generators/rails/templates/api_controller.rb
+++ b/lib/generators/rails/templates/api_controller.rb
@@ -48,13 +48,19 @@ class <%= controller_class_name %>Controller < ApplicationController
   private
     # Use callbacks to share common setup or constraints between actions.
     def set_<%= singular_table_name %>
+      <%- if Rails::VERSION::MAJOR >= 8 -%>
+      @<%= singular_table_name %> = <%= orm_class.find(class_name, "params.expect(:id)") %>
+      <%- else -%>
       @<%= singular_table_name %> = <%= orm_class.find(class_name, "params[:id]") %>
+      <%- end -%>
     end
 
     # Only allow a list of trusted parameters through.
     def <%= "#{singular_table_name}_params" %>
       <%- if attributes_names.empty? -%>
       params.fetch(<%= ":#{singular_table_name}" %>, {})
+      <%- elsif Rails::VERSION::MAJOR >= 8 -%>
+      params.expect(<%= singular_table_name %>: [ <%= permitted_params %> ])
       <%- else -%>
       params.require(<%= ":#{singular_table_name}" %>).permit(<%= permitted_params %>)
       <%- end -%>
--- a/test/scaffold_api_controller_generator_test.rb
+++ b/test/scaffold_api_controller_generator_test.rb
@@ -38,8 +38,17 @@ if Rails::VERSION::MAJOR > 4
           assert_match %r{@post\.destroy}, m
         end
 
+        assert_match %r{def set_post}, content
+        if Rails::VERSION::MAJOR >= 8
+          assert_match %r{params\.expect\(:id\)}, content
+        else
+          assert_match %r{params\[:id\]}, content
+        end
+
         assert_match %r{def post_params}, content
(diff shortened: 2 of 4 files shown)


MatheusRich:
@jeromedalbert add a Changelog entry, please

lib/generators/rails/templates/api_controller.rb
```diff
@@ -48,16 +48,28 @@ def destroy
...
+      <%- else -%>
       @<%= singular_table_name %> = <%= orm_class.find(class_name, "params[:id]") %>
+      <%- end -%>
     end
 
     # Only allow a list of trusted parameters through.
     def <%= "#{singular_table_name}_params" %>
+      <%- if Rails::VERSION::MAJOR >= 8 -%>
+      <%- if attributes_names.empty? -%>
+      params.expect(<%= singular_table_name %>: {})
```
martinemde, line 62: If we replace `params.fetch` like this it changes the behavior. Now, even though there are no attributes expected, you must send at least one attribute or you'll get a 400 error.

I suggest changing only the case with attributes and leaving this one alone.

test/scaffold_api_controller_generator_test.rb
```diff
@@ -52,7 +61,11 @@ class ScaffoldApiControllerGeneratorTest < Rails::Generators::TestCase
 
       assert_file 'app/controllers/posts_controller.rb' do |content|
         assert_match %r{def post_params}, content
-        assert_match %r{params\.fetch\(:post, \{\}\)}, content
+        if Rails::VERSION::MAJOR >= 8
+          assert_match %r{params\.expect\(post: \{\}\)}, content
+        else
+          assert_match %r{params\.fetch\(:post, \{\}\)}, content
+        end
```
martinemde, line 68: Same here, just revert this change.

martinemde (commented):
Thanks for following up with this. I think we should not change the `fetch` cases. Otherwise looks good (with changelog as mentioned).

lib/generators/rails/templates/api_controller.rb
jeromedalbert: @martinemde This makes sense, should the same be reverted from your PR on the Rails side?

https://github.com/rails/rails/pull/51674/files#diff-b20602dd9d67c64034125da9a3708746393dd155baa893f3bae0801dfb0efe11R51

and

https://github.com/rails/rails/pull/51674/files#diff-78fc500caeccee5a125dc41b391dec245fc90864a13ba44ef5ce009887438cd7R52

martinemde: You're exactly right. I submitted rails/rails#52932 to fix it. Thanks for noticing that I was in violation of my own recommendation. Thanks!

martinemde: This was the purpose of `params.allow`. Please someone from rails core let me know if that's worth a second, more focused attempt and I can extract the original code back out.

jeromedalbert:
@MatheusRich This project does not seem to have a Changelog file. 😅

I may be wrong but after looking at this:

https://github.com/rails/jbuilder/blob/acf44b86eebeb229dea651f2a8ec4bdd64a5328c/jbuilder.gemspec#L26

It looks like changelogs are made directly from GitHub at release time: https://github.com/rails/jbuilder/releases

Issue #363, opened 2016-10-12
ignore_nil! doesn't work for nil object values within array


aubergene:
This test demonstrates my issue

``` ruby
test 'ignore_nil! within array' do
  result = jbuild do |json|
    json.ignore_nil! true
    json.array! [{ name: 'Bob', dne: nil }]
  end

  assert_equal ['name'], result[0].keys

  result = jbuild do |json|
    json.ignore_nil! false
    json.array! [{ name: 'Bob', dne: nil }]
  end

  assert_equal ['name', 'dne'], result[0].keys
end
```

I'm afraid I couldn't work out how to fix it. Additionally I noticed that the test shows a symbol key is returned instead of a key, which might be related to the issue.

```
  diff:
  - ["name", "dne"]
  ?  ^    -  ^   -
  + [:name, :dne]
```

PR #580, opened 2025-02-07
Add support for ignoring nil values in Jbuilder scope

moritonal:
Okay, this isn't clear to merge yet. I'm doing this to see how the Actions run because I think there's currently a bug in it's building.


moritonal:
Okay, Ruby isn't my primary language so any advice is welcome on how to improve this. But it should fix up the issue whereby nested `null`'s aren't processed correctly.

moritonal:
Attempts to solve #363

PR #594, opened 2025-05-28
Cleanup project for Rails 7+ support

moberegger:
Recommended to review with white space off - most of the changes are adjustments to indentation as a result of removing several `if` conditions against versions of Rails. Also recommended to review commit by commit.

Support was dropped for EOL Ruby and Rails versions back in September in https://github.com/rails/jbuilder/pull/570. Part of the motivation to do this was to address some test issues in [another PR](https://github.com/rails/jbuilder/pull/569). Since then, nothing below Ruby 3.0 and Rails 7.0 run as part of the test matrix. Given that the project no longer runs validation against older versions of Ruby and Rails, compatibility can no longer be guaranteed. I think it makes sense to update the project to reflect this reality.

A summary of the changes:
- `gemspec` has been updated to specify Ruby 3.0 is the minimum version, as well as specifying 7.0 as the minimum versions for both `activesupport` and `actionview`, since this is what the test matrix actually runs against.
- Test cases have been cleaned up to remove some redundant `if` conditions against Ruby/Rails versions that would always be `true` in the test matrix. Likewise, conditions and assertions for older versions have been removed since they would no longer be running under the test matrix.
- The railtie has been tidied up to assume Rails 7+
- The `CollectionRenderer` has been refactored to trust that `::ActionView::CollectionRenderer` exists, which is the case for Rails 7+. This allows us to remove the `::ActionView::PartialRenderer` shim which also allows us to remove the branch of logic in `JbuilderTemplate` where `CollectionRenderer.supported?` is `false`, which will never be the case anymore in Rails 7+.

### DIFF
--- a/lib/jbuilder/railtie.rb
+++ b/lib/jbuilder/railtie.rb
@@ -9,28 +9,24 @@ class Jbuilder
         require 'jbuilder/jbuilder_dependency_tracker'
       end
 
-      if Rails::VERSION::MAJOR >= 5
-        module ::ActionController
-          module ApiRendering
-            include ActionView::Rendering
-          end
+      module ::ActionController
+        module ApiRendering
+          include ActionView::Rendering
         end
+      end
 
-        ActiveSupport.on_load :action_controller do
-          if name == 'ActionController::API'
-            include ActionController::Helpers
-            include ActionController::ImplicitRender
-          end
+      ActiveSupport.on_load :action_controller do
+        if name == 'ActionController::API'
+          include ActionController::Helpers
+          include ActionController::ImplicitRender
         end
       end
     end
 
-    if Rails::VERSION::MAJOR >= 4
-      generators do |app|
-        Rails::Generators.configure! app.config.generators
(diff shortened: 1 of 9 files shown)


moberegger:
@rafaelfranca Any concerns here? Is my reasoning sound, or is this change considered too risky?

lib/jbuilder/railtie.rb
```diff
@@ -9,28 +9,24 @@ class Railtie < ::Rails::Railtie
...
-          if name == 'ActionController::API'
-            include ActionController::Helpers
-            include ActionController::ImplicitRender
-          end
+      ActiveSupport.on_load :action_controller do
+        if name == 'ActionController::API'
+          include ActionController::Helpers
+          include ActionController::ImplicitRender
         end
       end
```
rafaelfranca, line 23:
```suggestion
      ActiveSupport.on_load :action_controller_api do
        include ActionController::Helpers
        include ActionController::ImplicitRender
      end
```

PR #606, opened 2025-11-14
fix: update overidden signature to match upstream

ThomasCrambert:
Since https://github.com/rails/rails/commit/107ee64909b978476b318e36f317cfa2a747d1ad, the signature of `collection_with_template` has been updated but the override in `lib/jbuilder/collection_renderer.rb` is still using the old one, resulting in `ActionView::Template::Error: wrong number of arguments (given 5, expected 4)`


ThomasCrambert:
Fixed thanks to https://github.com/rails/rails/pull/56140

PR #610, opened 2026-01-12
Performance improvement for array and hash

mayur-kambariya:


rafaelfranca:
Can you provide benchmark and results for those performance improvements? Otherwise we can't merge

PR #613, opened 2026-04-03
fix: Preserve locals when rendering inline partial for object

moberegger:
Ran into an odd bug. Given a template like

```ruby
# locals (json:, foo:, my_local: 'default')

json.extract! foo, :id, :name
json.myLocal my_local
```

Rendering to a collection inline like

```ruby
json.foos, @foos, partial: 'foos/foo', as: :foo, my_local: 'custom'
```

Would correctly render with the value of `my_local` with the value of `'custom'`

```json
{
  "foos": [
    "id": 1,
    "name": "Test",
    "myLocal": "custom"
  ]
}
```

But rendering a single object _wasn't_ picking up the value for the local

```ruby
json.foo, @foo, partial: 'foos/foo', as: :foo, my_local: 'custom'
```

```json
{
  "foo": {
    "id": 1,
    "name": "Test",
    "myLocal": "default"
  }
}
```

If the partial _didn't_ have a default value set for strict locals

```ruby
# locals (json:, foo:, my_local:)

json.extract! foo, :id, :name
json.myLocal my_local
```

The following 

```ruby
json.foo, @foo, partial: 'foos/foo', as: :foo, my_local: 'custom'
```

would raise

```
ActionView::Template::Error: missing local: :my_local for app/views/foos/_foo.json.jbuilder...
```

The reason is because this render path in `_set_inline_partial` would set a key for `locals` before the render is done in `_render_partial_with_options`.  `_render_partial_with_options` attempts to pluck key args for `locals`, but since the key is already there, they are effectively ignored.

This _may_ have been a regression introduced by #591 which made it no longer `reverse_merge!` on the `locals` key. I haven't verified this, though.

The fix was to not have `_set_inline_partial` set `locals`, and instead defer that logic to `_render_partial_with_options`.

I added some extra tests cases to cover this.
- A test for rendering an object to an inline partial (there were no tests for this)
- A test for rendering an object to an inline partial with locals
- A test for rendering a collection to an inline partial with locals

I also updated the docs to include an example for rendering an object to an inline partial. This is supported by the library, but was seemingly undocumented. Also documented rendering to partials inline with other locals, which was also supported by the library but seemingly undocumented.

As an aside: With some small refactoring, I think we can drop the need for the intermediary `_set_inline_partial` method. It doesn't do too much and I think the coordination of `locals` can all be handled within `_render_partial_with_options`. This would save a bit on indirection and also align everything to the same render path. Outside of the scope of this PR, but perhaps something I can try later.

### DIFF
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -174,13 +174,9 @@ class JbuilderTemplate < Jbuilder
         array!
       end
     else
-      _render_partial options
+      options[:locals][:json] = self
+      @context.render options
     end
-  end
-
-  def _render_partial(options)
-    options[:locals][:json] = self
-    @context.render options
   end
 
   def _cache_fragment_for(key, options, &block)
@@ -241,7 +237,7 @@ class JbuilderTemplate < Jbuilder
       end
     else
       _scope do
-        options[:locals] = { options[:as] => object }
+        options[options[:as]] = object
         _render_partial_with_options options
       end
     end
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ json.array! @comments do |comment|
   end
 end
 
-# => [ { "body": "great post...", "author": { "first_name": "Joe", "last_name": "Bloe" }} ]
+# => [ { "body": "great post...", "author": { "first_name": "Joe", "last_name": "Blow" }} ]
 ```
 
 ## Array Attributes
@@ -209,6 +209,13 @@ the partial.
 json.partial! 'comments/comments', comments: @message.comments
 ```
 
+You can also render an object to a partial inline under a key.
+
+```ruby
+json.post @post, partial: 'posts/post', as: :post
+# => { "post": { "title": "Hello World!", "author": { "name": "David" } } }
+```
+
 It's also possible to render collections of partials:
 
 ```ruby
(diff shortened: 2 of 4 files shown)


lib/jbuilder/jbuilder_template.rb
```diff
@@ -241,7 +237,7 @@ def _set_inline_partial(name, object, options)
       end
     else
       _scope do
-        options[:locals] = { options[:as] => object }
+        options[options[:as]] = object
```
moberegger, line 240: This is the crux of the fix.

lib/jbuilder/jbuilder_template.rb
```diff
@@ -174,15 +174,11 @@ def _render_partial_with_options(options)
...
-      _render_partial options
+      options[:locals][:json] = self
+      @context.render options
     end
   end
 
-  def _render_partial(options)
-    options[:locals][:json] = self
-    @context.render options
-  end
```
moberegger, line 184: Since this only had one call site...

lib/jbuilder/jbuilder_template.rb
```diff
@@ -174,15 +174,11 @@ def _render_partial_with_options(options)
         array!
       end
     else
-      _render_partial options
+      options[:locals][:json] = self
+      @context.render options
```
moberegger, line 178: ... I just inlined it here.

PR #620, opened 2026-07-16
Register dependency tracker so nested partials invalidate parent caches

rosa:
We ran into a caching issue at Basecamp with a template that looked like this: 

```
json.cache! ... do
  # ...
  
    json.people people, partial: "people/person", as: :person
end
```

We changed the `people/_person.json.jbuilder` template and deployed, yet responses using that template kept returning a stale cached version of the person. 

We traced this down to nothing registering Jbuilder's dependency tracker with Action View. Its railtie requires it, but nothing registers it with `ActionView::DependencyTracker`. Without a registered tracker, [`find_dependencies`](https://github.com/rails/rails/blob/f011d1218bdd77857f30ce3964eef186f0f14de5/actionview/lib/action_view/digestor.rb#L55) returns `[]` for every `.jbuilder` template, so the digestor sees no nested-partial dependencies: a `json.cache!` fragment that renders inner partials never mixes those partials' digests into its own cache key. Editing an inner partial then leaves outer fragments serving stale cached JSON until they happen to invalidate for some unrelated reason. ERB doesn't have this problem because [Action View registers `ERBTracker`](https://github.com/rails/rails/blob/f011d1218bdd77857f30ce3964eef186f0f14de5/actionview/lib/action_view.rb#L103).

This is a regression from #504, which rewrote the tracker as a standalone class and dropped the `register_tracker :jbuilder` call (and the `ERBTracker` inheritance that used to supply it) without re-adding it. It has shipped unregistered since 2.12.0.

Register the tracker from the same `on_load(:action_view)` block that registers the template handler, and declare `supports_view_paths?` so `register_tracker passes` the view paths through to the tracker. Otherwise, Rails wraps it in a lambda that drops them and the tracker's own wildcard (`# Template Dependency: foo/*`) resolution silently returns nothing.

### DIFF
--- a/lib/jbuilder/jbuilder_dependency_tracker.rb
+++ b/lib/jbuilder/jbuilder_dependency_tracker.rb
@@ -25,6 +25,10 @@ class Jbuilder::DependencyTracker
         (['"])([^'"]+)\1            # quoted value
       /x
 
+  def self.supports_view_paths?
+    true
+  end
+
   def self.call(name, template, view_paths = nil)
     new(name, template, view_paths).dependencies
   end
--- a/lib/jbuilder/railtie.rb
+++ b/lib/jbuilder/railtie.rb
@@ -8,7 +8,10 @@ class Jbuilder
     initializer :jbuilder do
       ActiveSupport.on_load :action_view do
         ActionView::Template.register_template_handler :jbuilder, JbuilderHandler
+
+        require 'action_view/dependency_tracker'
         require 'jbuilder/jbuilder_dependency_tracker'
+        ActionView::DependencyTracker.register_tracker :jbuilder, Jbuilder::DependencyTracker
       end
 
       module ::ActionController
--- a/test/jbuilder_dependency_tracker_test.rb
+++ b/test/jbuilder_dependency_tracker_test.rb
@@ -8,6 +8,22 @@ class FakeTemplate
     end
 end
 
+FakeResolvedPath = Struct.new(:prefix, :virtual_path) do
+    def to_s
+      virtual_path
+    end
+end
+
+class FakeViewPath
+    def initialize(*template_paths)
+      @template_paths = template_paths
+    end
+
+    def all_template_paths
+      @template_paths
+    end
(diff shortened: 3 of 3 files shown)


seanpdoyle (approved):

seanpdoyle:
While I believe that this is a needed improvement that restores prior behavior, there _have_ been several releases since the regression was introduced. Should integration with the dependency tracker be guarded by an engine configuration value, in case applications that depend on intermediate versions have relied on the *absence* of the dependency tracker integration?

rosa:
@seanpdoyle, thanks a lot for checking this! That's a good point 🤔 I think it'd be a bit strange to rely on stale cached content, where you might have a version of a template somewhere and another version somewhere else... maybe with different attributes and values... But it's possible! I did wonder why this had gone unnoticed for so long. I mean, we had this broken in our app since 2024. The reason was that we don't have that many examples of partial jbuilder templates inside a cached fragment, where we also change the partial, and the specific partial that triggered the bug for us changed very rarely. I suppose this is normal for folks as well?

Personally, I wouldn't add a configuration setting for a bug like this because I can't imagine how you rely on this behaviour in particular, but maybe you have a case in mind?

etiennebarrie:
Sorry I had missed that but came to the same fix. Agreed that it's a bug not worth adding a configuration for (which has a risk of being missed, then the same problem will appear at the next major/whenever the configuration is dropped).

If we want to be helpful, we could mention a way to disable it:
```ruby
# config/initializers/jbuilder.rb
ActiveSupport.on_load(:action_view) do
  ActionView::DependencyTracker.remove_tracker(ActionView::Template.handler_for_extension(:jbuilder))
end
```

And since template digests are stable, migrating a cache is doable, you "just" need to find the current digest and the one including dependencies (from logs in development), and then do a bunch of cache get/set. Then remove the initializer, and then migrate one partial at a time by changing them/dropping the cache for those digests. But that should only be necessary for applications that truly depend on their cache being warm to perform adequately.

Issue #250, opened 2015-02-21
Best approach for common API response


mikeys:
I'm building a Rails based API, and I'd like to use JBuilder for rendering the JSON responses.

The standard response for all calls for my API should be as follows:
```js
{
  status: "ok",
  error: "",
  data: {
    // Changes based on specified action, should be yielded
  }
}
```

Since I couldn't see any support for layouts in JBuilder, what's the best practice for supporting the above response?

Is there any way to do the above without adding 'status' and 'error' partials for every jbuilder view?

rwz:
The best approach would probably be a view helper.

```ruby
def json_data_response
  json = local_assigns[:json]
  json.status "ok"
  json.error ""
  json.data &block
end
```

so you can use it in your Jbuilder views like this:
```ruby
json_data_response do
  json.foo "bar"
end 
```

to get
```javascript
{
  "status": "ok",
  "error": "",
  "data": { "foo": "bar" }
}
```

mikeys:
That makes sense, thanks :)

lneffa:
I'm having problems implementing this approach, I have placed the json_data_response helper inside application_helper.rb, and inside a jbuilder view I can access the helper, but local_assigns is undefined and any attempt to use defined? or access json module is undefined. 

What steps am I missing? Thank you for the help

Issue #172, opened 2014-03-13
Ability to use a layout with yield without parsing and encoding two times


alex88:
Actually to use a json template it's needed to `JSON.parse(yield)` which needs a re-parse of the encoded content and a re-encoding to the final json.

It would be nice to be able to skip this step to improve performances.

jplethier:
I need to use a layout just like this, is there any news or plans to implement a way to not need to re-parse and re-encoding json again?

Agent47DarkSoul:
@jplethier I tried creating the application.json.jbuilder and rails does use it as a layout and tries to render it. Unfortunately it raises an error regarding stored values in lib/action_view/flows.rb:13. I am trying to figure out what the issue is, will post an update when I do.

amw:
I'm not sure why the opposition to include functionality like this. I think wrapping response data in a shared metadata structure is commonplace in API design. See `data`, `errors` or `meta` keys in JSON-API specification: http://jsonapi.org/format/.

Not a very pretty solution, but I managed to create my general response layout using json.erb and just pasting literal results of `yield` in a proper position. Much better than doing an extra round of `JSON.parse` & `JSON.generate`.

app/controllers/api/base_controller.rb:
```
class Api::BaseController < ActionController::Base
  layout "api.v1"
end
```

app/views/layouts/api.v1.json.erb:
```
{
<% if @api_errors.present? %>
  "errors": <%= JSON.dump @api_errors %>,
<% else %>
  "data": <%= yield %>,
<% end %>
  "meta": <%= JSON.dump @api_meta %>
}
```

paulwalker:
It's pretty remarkable actually that Rails 5 (with all it's special emphasis on building JSON APIs) would launch without the ability to easily render the JSON-API standard, of which @wycats himself is a primary editor. All because dhh doesn't believe in metadata in api responses?

Richard-Degenne:
Is there any news for this feature?

seanpdoyle:
I've opened https://github.com/rails/jbuilder/pull/623 in an effort to implement this feature.

PR #623, opened 2026-09-04
Support integrating with Action View layouts

seanpdoyle:
Related to https://github.com/rails/jbuilder/issues/250 
Related to https://github.com/rails/jbuilder/issues/172

Add support for Rails' built-in layout templates defined in `app/views/layouts/`. Layouts can be used to extract properties or structures that are shared across responses.

For example, a `PostsController` that inherits from `ApplicationController` will render its templates declared in `app/views/posts/` *into* the layout declared in
`app/views/layouts/application.json.jbuilder`:

```ruby
# app/views/layouts/application.json.jbuilder
json.merge! yield
json.servedAt Time.current

# app/views/posts/index.json.jbuilder
json.partial! 'posts/post', collection: @posts, as: :post
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,25 @@ json.partial! 'sub_template', locals: { user: user }
 json.partial! 'sub_template', user: user
 ```
 
+## Layouts
+
+You can rely on Rails' built-in support for layout templates defined in
+`app/views/layouts/`. Layouts can be used to extract properties or structures
+that are shared across responses.
+
+For example, a `PostsController` that inherits from `ApplicationController` will
+render its templates declared in `app/views/posts/` *into* the layout declared
+in `app/views/layouts/application.json.jbuilder`:
+
+```ruby
+# app/views/layouts/application.json.jbuilder
+json.merge! yield
+json.servedAt Time.current
+
+# app/views/posts/index.json.jbuilder
+json.partial! 'posts/post', collection: @posts, as: :post
+```
+
 ## Null Values
 
 You can explicitly make Jbuilder object return null if you want:
--- a/lib/jbuilder/jbuilder_template.rb
+++ b/lib/jbuilder/jbuilder_template.rb
@@ -2,6 +2,7 @@
 
 require 'jbuilder/jbuilder'
 require 'jbuilder/collection_renderer'
+require 'jbuilder/rendering_helper'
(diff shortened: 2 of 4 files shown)


malomalo:
@seanpdoyle I maintain [TurboStreamer](https://github.com/malomalo/turbostreamer) it might be a better fit for you. It's a Jbuilder-ish DSL over streaming encoders (Oj/Yajl), and supports layouts and improved caching performance.

A layout calls `json.yield!` wherever the template's JSON belongs:

```ruby
# app/views/layouts/application.json.streamer
json.object! do
  json.meta do
    json.object! { json.version 1 }
  end
  json.key! :data
  json.yield!
end

# app/views/posts/index.json.streamer
json.array! @posts, :id, :title

# => { "meta": { "version": 1 }, "data": [ { "id": 1, "title": "..." } ] }
```

The layout and the template share one builder, so the template writes straight into the same stream at the point it's yielded — nothing is encoded to a string and re-parsed, which also means the template's JSON can land at a key or inside an array rather than only merging into the root. This also enables streaming support.

More in the [README](https://github.com/malomalo/turbostreamer#layouts). Happy to answer questions if you give it a look.

seanpdoyle:
@malomalo That seems like a compelling project, thanks for the recommendation! 

I'm sure that there's plenty of innovation out there worth trying. The motivation behind this PR is less about innovating, and more about bringing `jbuilder` more in line with existing Action View and Rails conventions. Since `jbuilder` is part of most `rails new`-generated Gemfiles, I think many would benefit from any improvements.

> supports layouts and improved caching performance.

That's exciting! Are there any opportunities to upstream performance improvements back into `jbuilder`?

malomalo:
> That's exciting! Are there any opportunities to upstream performance improvements back into `jbuilder`?

@seanpdoyle I tried for a while with my fork to get jBuilder off hashes at least partially but if i recall correctly it's impossible due to syntax and needing to know how to insert the string. I forget the concrete example convinced me it wasn't possible.

I just remember your coworkers did something [similar](https://thoughtbot.com/blog/introducing-props-template) recently they might have and idea.

*redacting/editing comment - I jumped to conclusions with a few things in the earlier comment and props-template although similar on first look approaches some things differently

